PR #5, opened 2016-07-20
Let devicemapper work on stable rust

cuviper:
Serde's serialization traits can be "derived" in a build script using
serde_codegen.  This patch follows serde's example to make device.rs.in
which can be built either way, with serde_macros or serde_codegen.

Clippy is nice to have, but entirely optional.

The default feature set is now compatible with stable rust.  To enable
both serde_macros and clippy features at once, use:
     cargo build --no-default-features --features nightly

Signed-off-by: Josh Stone <jistone@redhat.com>

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,2 +1,11 @@
 language: rust
-rust: nightly
+rust:
+  - stable
+  - beta
+  - nightly
+script:
+  - cargo build --verbose && cargo test --verbose
+  - |
+    [ $TRAVIS_RUST_VERSION != nightly ] ||
+    ( cargo build --verbose --no-default-features --features nightly &&
+      cargo test --verbose --no-default-features --features nightly )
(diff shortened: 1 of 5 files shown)


cuviper:
However, in trying Froyo against this, I just discovered that `quasi_macros` fails to compile.  There seems to be some bad feature interaction when `serde_codegen` and `serde_macros` are both active in a build, per this default-stable `devicemapper` and still-nightly `froyo`, respectively.

It does work if I force everyone to nightly, like so in Froyo's `Cargo.toml`:

```
[dependencies.devicemapper]
version = "0.3.0"
default-features = false
features = ["nightly"]
```

But if Froyo starts trying to also go either way, I fear it will get messy...

Perhaps it would be better to commit fully to the stable `serde_codegen`, rather than trying to straddle it?  With a semver bump here, to give Froyo a chance to catch up.  Clippy can still be an optional nightly-only feature.  What do you think?

agrover:
Yeah it's probably time to move in this direction.

cuviper:
Closing in favor of #6.  The `Device` serialization isn't really necessary.

PR #12, opened 2017-03-29
move LinearDev to DM, add logging, move IEC constants

trgill:
Move LinearDev from Stratis to devicemapper-rs.

Updated LinearDev::new() to take a vector of struct Segment.
Changed LinearDev functions to return DmResult
Moved wait_for_dm to DM.
Use DM::wait_for_dm() for lineardev.

Misc:
Moved IEC constants 
Added logging for debug!() used in LinearDev

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,3 +22,5 @@ clippy = { version = "0", optional = true }
 newtype_derive = "0.1"
 custom_derive = "0.1"
 serde = "0"
+log = "0.3"
+env_logger="0.3.5"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -166,8 +166,8 @@ pub struct Device {
 }
 
 impl Device {
-    /// Returns the path in `/dev` that corresponds with the device number.
-    pub fn path(&self) -> Option<PathBuf> {
+    /// Returns the path in `/dev` that corresponds with the device number/devnode.
+    pub fn devnode(&self) -> Option<PathBuf> {
         let f = File::open("/proc/partitions").expect("Could not open /proc/partitions");
 
         let reader = BufReader::new(f);
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -57,13 +57,24 @@ extern crate serde;
 #[macro_use]
 extern crate bitflags;
 
+#[macro_use]
+extern crate log;
+
 #[allow(dead_code, non_camel_case_types)]
 mod dm_ioctl;
-mod util;
-/// Module for basic types (Bytes, Sectors, DataBlocks)
+/// public utilities
+pub mod util;
+/// basic types (Bytes, Sectors, DataBlocks)
 pub mod types;
-/// Module for shared constants
+/// shared constants
 pub mod consts;
+/// functions to create continuous linear space given device segments
+pub mod lineardev;
+/// struct to represent a location, offset and size of a set of disk sectors
+pub mod segment;
+/// return results container
+pub mod result;
+
 
 use std::fs::File;
 use std::io;
@@ -78,6 +89,8 @@ use std::collections::BTreeSet;
 use std::os::unix::fs::MetadataExt;
 use std::cmp;
 use std::borrow::Borrow;
+use std::thread;
+use std::time::Duration;
 
 use nix::sys::ioctl::ioctl as nix_ioctl;
 use nix::sys::ioctl::libc::c_ulong;
@@ -225,10 +238,11 @@ pub fn dev_majors() -> BTreeSet<u32> {
 
     let reader = BufReader::new(f);
 
-    for line in reader.lines()
-        .filter_map(|x| x.ok())
-        .skip_while(|x| x != "Block devices:")
-        .skip(1) {
+    for line in reader
+            .lines()
+            .filter_map(|x| x.ok())
+            .skip_while(|x| x != "Block devices:")
+            .skip(1) {
         let spl: Vec<_> = line.split_whitespace().collect();
 
         if spl[1] == "device-mapper" {
@@ -316,6 +330,12 @@ impl DM {
         Ok(DM { file: try!(File::open(DM_CTL_PATH)) })
     }
 
+    /// The /dev/mapper/<name> device is not immediately available for use.
+    /// TODO: Implement wait for event or poll.
+    pub fn wait_for_dm() {
+        thread::sleep(Duration::from_millis(500))
+    }
+
     fn initialize_hdr(hdr: &mut dmi::Struct_dm_ioctl, flags: DmFlags) -> () {
         hdr.version[0] = DM_VERSION_MAJOR;
         hdr.version[1] = DM_VERSION_MINOR;
@@ -376,12 +396,16 @@ impl DM {
         let op = iorw!(DM_IOCTL, ioctl, size_of::<dmi::Struct_dm_ioctl>()) as c_ulong;
         loop {
             if let Err(_) = unsafe {
-                convert_ioctl_res!(nix_ioctl(self.file.as_raw_fd(), op, v.as_mut_ptr()))
-            } {
+                   convert_ioctl_res!(nix_ioctl(self.file.as_raw_fd(), op, v.as_mut_ptr()))
+               } {
                 return Err((Error::last_os_error()));
             }
 
-            let hdr = unsafe { (v.as_mut_ptr() as *mut dmi::Struct_dm_ioctl).as_mut().unwrap() };
+            let hdr = unsafe {
+                (v.as_mut_ptr() as *mut dmi::Struct_dm_ioctl)
+                    .as_mut()
+                    .unwrap()
+            };
 
             if (hdr.flags & DM_BUFFER_FULL.bits) == 0 {
                 break;
@@ -392,7 +416,11 @@ impl DM {
             hdr.data_size = v.len() as u32;
         }
 
-        let hdr = unsafe { (v.as_mut_ptr() as *mut dmi::Struct_dm_ioctl).as_mut().unwrap() };
+        let hdr = unsafe {
+            (v.as_mut_ptr() as *mut dmi::Struct_dm_ioctl)
+                .as_mut()
+                .unwrap()
+        };
 
         // hdr possibly modified so copy back
         hdr_slc.clone_from_slice(&v[..hdr.data_start as usize]);
@@ -448,7 +476,9 @@ impl DM {
 
             loop {
                 let device = unsafe {
-                    (result.as_ptr() as *const dmi::Struct_dm_name_list).as_ref().unwrap()
+                    (result.as_ptr() as *const dmi::Struct_dm_name_list)
+                        .as_ref()
+                        .unwrap()
                 };
 
                 let slc = slice_to_null(&result[size_of::<dmi::Struct_dm_name_list>()..])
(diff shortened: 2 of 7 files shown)


trgill:
@mulkieran @agrover review please....

src/lineardev.rs
```diff
@@ -0,0 +1,109 @@
...
+            Ok(bytes) => return Ok(bytes),
+            Err(e) => {
+                let error_message = format!("{:?} for device {:?}", e, self.path());
+                return Err(DmError::Dm(InternalError(error_message.into())));
+            }
+        };
+    }
+    /// path of the device
+    pub fn path(&self) -> DmResult<PathBuf> {
+        match self.dev_info.device().path() {
```
mulkieran: ```self.dev_info.device().path().ok_or(DmError::Dm(InternalError("No path associated with dev_info".into()))``` is more idiomatic.

src/lineardev.rs
```diff
@@ -0,0 +1,109 @@
...
+                        "linear".to_owned(),
+                        format!("{} {}", dstr, *start));
+            debug!("dmtable line : {:?}", line);
+            table.push(line);
+            logical_start_sector = logical_start_sector + length;
+        }
+
+        table
+    }
+    /// name of the linear device
```
mulkieran: I'm not really sure this ```name()``` method is really necessary. It is only used in Debug...something else might be just as useful in Debug.

src/segment.rs
```diff
@@ -0,0 +1,36 @@
...
+    /// Create a new Segment with given atributes
+    pub fn new(devnode: PathBuf, start: u64, length: u64) -> Segment {
+        Segment {
+            devnode: devnode,
+            start: Sectors(start),
+            length: Sectors(length),
+        }
+    }
+    /// List the available-for-upper-layer-use range in this blockdev.
+    pub fn range(&self) -> (Sectors, Sectors) {
```
mulkieran: Might be reasonable to come up with a more descriptive name, e.g., ```free_range``` as in chickens.

src/lineardev.rs
```diff
@@ -0,0 +1,109 @@
...
+    pub fn name(&self) -> &str {
+        self.dev_info.name()
+    }
+    /// return the total size of the linear device
+    pub fn size(&self) -> DmResult<Bytes> {
+
+        let blockdev_path = try!(self.path());
+        let f = match File::open(blockdev_path) {
```
mulkieran: Perhaps these two are try!-able in their new context?

mulkieran:
IEM to IEC wherever appropriate in commit messages.

src/lineardev.rs
```diff
@@ -0,0 +1,109 @@
...
+use std::path::PathBuf;
+
+use util::blkdev_size;
+use result::{DmResult, DmError, InternalError};
+use types::{Bytes, Sectors};
+
+/// A DM construct of combined Segments
+pub struct LinearDev {
+    /// Name that as device reference - /dev/mapper/<name>
+    name: String,
```
trgill: It isn't a complete path, right?  It is just the <name> part of  /dev/mapper/<name>.

src/lineardev.rs
```diff
@@ -0,0 +1,109 @@
...
+            None => {
+                return Err(DmError::Dm(InternalError("No path associated with dev_info".into())))
+            }
+        }
+    }
+
+    /// Remove the device from DM
+    pub fn teardown(&self, dm: &DM) -> DmResult<()> {
+        try!(dm.device_remove(&DevId::Name(&self.name), DmFlags::empty()));
+
```
mulkieran: Maybe you can lose the whole line and the ```try!``` ?

src/lineardev.rs
mulkieran: I see, the comment confused me. What about comment like:

```/// Device mapper file name - /dev/mapper/<name>```

It must have a more specific description/name, since the devicemapper people must think it is different from the value returned by the ```name()``` function below.

File: src/segment.rs [hunk 7]
```diff
@@ -0,0 +1,36 @@
...
+        (self.start, self.length)
+    }
+
+    /// Get the "x:y" device string for this blockdev
+    pub fn dstr(&self) -> String {
```
trgill: Ok.   This is a suggestion to change the name?  How about dev_str()?

src/segment.rs
trgill: It is a just a segment that could represent a free range, but could also be a reserved range.

src/segment.rs
mulkieran: This devicemapper library uses "dstr" to mean what the comment above says. I honestly don't know why that name was chosen. The problem is that the code doesn't seem to match the header comment or the understood meaning of "dstr" in devicemapper.

src/segment.rs
mulkieran: You might want to change the header comment to be a bit more generic...

src/segment.rs
agrover: Yeah, a "dstr" is just a string with the major and minor numbers separated by a colon. This is used frequently in dm target parameters. This is easy to generate if you have a `Device` struct since that has maj and min as members (and implements `.dstr()`).

Thus if you had a `Device` in `Segment` instead of a `PathBuf`, this would be easy. Perhaps that's the way to go? (You can get a `Path` from a `Device`, and you can get a `Device` from a &str (which you can get from a `Path`))

mulkieran: @agrover Should Device implement ```AsString``` for this, instead of a specially named method? It does seem like the canonical string for a Device.

agrover: Can't find AsString, you mean `ToString`? If so then maybe. Or maybe we implement `fmt::Display` and that automatically gives us `ToString`.

I think this is worthwhile but `dstr()` removal is a whole nother thing that we shouldn't hold up this PR for. Can we just remove this function for now? I don't actually see `Segment::dstr` being used anywhere.

mulkieran: I agree that dstr() -> ToString impl can be deferred until whenever. But ```dstr()``` is used, lineardev.rs: line 57, about.

src/segment.rs
trgill: Ok - updated to : 
    /// Return the start and length sectors

src/segment.rs
trgill: I think changing devnode to a Device is a good idea.  

I do need dstr() to create the tables from Segments, so I'll make that change.

src/lineardev.rs
trgill: Ok - I updated the comment to your suggestion and opened https://github.com/stratis-storage/stratisd/issues/295 because we've got other problems in this area.

src/lineardev.rs
```diff
@@ -0,0 +1,109 @@
...
+    }
+}
+
+/// Use DM to concatenate a set of blockdevs together into a
+/// /dev/mapper/xxx block device of continuous sectors.
+impl LinearDev {
+    /// Construct a new block device by concatenating the given block_devs
+    /// into linear space.  Use DM to reserve enough space for the stratis
+    /// metadata on each DmDev.
+    pub fn new(name: &str, dm: &DM, block_devs: &Vec<&Segment>) -> DmResult<LinearDev> {
```
trgill: Ok

src/lineardev.rs
```diff
@@ -0,0 +1,109 @@
...
+        DM::wait_for_dm();
+        Ok(LinearDev {
+            name: name.to_owned(),
+            dev_info: dev_info,
+        })
+    }
+
+    /// Generate a Vec<> to be passed to DM.  The format of the Vec entries is:
+    /// <logical start sec> <length> "linear" /dev/xxx <start offset>
+    fn dm_table(block_devs: &Vec<&Segment>) -> Vec<(u64, u64, String, String)> {
```
trgill: Ok

src/lineardev.rs
trgill: This is the DM name - from the DeviceInfo struct.  It is very useful for debugging - If we build things on top of it - we will need it to construct the table.

src/lineardev.rs
trgill: Ok

src/lineardev.rs
trgill: Ok

src/lineardev.rs
trgill: Ok

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "devicemapper"
-version = "0.4.1"
+version = "0.4.2"
```
agrover: It may be best to increment the version number separately from functional PRs. This should avoid conflicts -- `master` is already 0.4.2 -- unless there's a reason that the other parts of this PR need this, but I can't think of what reasons there might be.

src/lib.rs
```diff
@@ -57,13 +57,24 @@ extern crate serde;
 #[macro_use]
 extern crate bitflags;
 
+#[macro_use]
```
agrover: please fix typo in this commit's summary

Cargo.toml
trgill: ok - sounds good.

agrover:
I'm still seeing "devicemapper" spelled wrong in 7ff4150

Completely trivial but please fix 😃 and then 👍to merge

src/lineardev.rs
mulkieran: Maybe the comment could be made a bit more descriptive?

src/segment.rs
```diff
@@ -0,0 +1,36 @@
...
+    length: Sectors,
+    devnode: Device,
+}
+
+
+impl Segment {
+    /// Create a new Segment with given atributes
+    pub fn new(devnode: Device, start: u64, length: u64) -> Segment {
+        Segment {
+            devnode: devnode,
```
mulkieran: This isn't a device node anymore...probably name should be changed to ```device```.

src/segment.rs
```diff
@@ -0,0 +1,36 @@
...
+#[derive(Debug)]
+pub struct Segment {
+    start: Sectors,
+    length: Sectors,
+    devnode: Device,
+}
+
+
+impl Segment {
+    /// Create a new Segment with given atributes
```
mulkieran: Typo on "atributes".

src/segment.rs
```diff
@@ -0,0 +1,36 @@
...
+        }
+    }
+    /// Return the start and length sectors
+    pub fn range(&self) -> (Sectors, Sectors) {
+        (self.start, self.length)
+    }
+
+    /// Get the "x:y" device string for this blockdev
+    pub fn dstr(&self) -> String {
+        format!("{:?}", self.devnode.dstr())
```
mulkieran: You don't need to use ```format!``` here.

src/lineardev.rs
```diff
@@ -0,0 +1,94 @@
...
+    /// return the total size of the linear device
+    pub fn size(&self) -> DmResult<Bytes> {
+
+        let blockdev_path = try!(self.path());
+        let f = try!(File::open(blockdev_path));
+        blkdev_size(&f)
+
+    }
+    /// path of the device
+    pub fn path(&self) -> DmResult<PathBuf> {
```
mulkieran: _This_ is the device node AFAIK. Can it be renamed to  ```devnode```?

src/lineardev.rs
```diff
@@ -0,0 +1,94 @@
...
+
+        table
+    }
+    /// name of the linear device
+    pub fn name(&self) -> &str {
+        self.dev_info.name()
+    }
+    /// return the total size of the linear device
+    pub fn size(&self) -> DmResult<Bytes> {
+
```
mulkieran: No harm to lose a few empty lines in this method.

src/lineardev.rs
trgill: It is calling Device::path().  The comment for that method is:

/// Returns the path in `/dev` that corresponds with the device number

src/segment.rs
trgill: Ok

src/segment.rs
trgill: Ok

src/lineardev.rs
trgill: Ok

src/lineardev.rs
mulkieran: I think Andy meant "device node", there.

mulkieran: We should take care to have precise, meaningful names by this 1.0 version.

src/lineardev.rs
trgill: Ok

trgill:
updated LinearDev::path() and Device::path() to ::devnode()

PR #18, opened 2017-04-13
Master infra

mulkieran:


mulkieran:
Expected to fail since warnings are treated as errors.

mulkieran:
@agrover review?

mulkieran:
@agrover the only difference is that instead of taking require missing docs off all, I take it off just for the one file.

trgill:
👍

PR #20, opened 2017-04-14
Add ThinPoolDev::status()

agrover:
Add a method that gets the DM status for the thinpool, and parses the
textual output. This code is lifted with minor modifications from Froyo.

### DIFF
--- a/src/thinpooldev.rs
+++ b/src/thinpooldev.rs
@@ -25,6 +25,42 @@ impl fmt::Debug for ThinPoolDev {
     }
 }
 
+#[derive(Debug, Clone, Copy)]
+/// Contains values indicating the thinpool's used vs total
+/// allocations for metadata and data blocks.
+pub struct ThinPoolBlockUsage {
+    /// The number of metadata blocks that are in use.
+    pub used_meta: u64,
+    /// The total number of metadata blocks available to the thinpool.
+    pub total_meta: u64,
+    /// The number of data blocks that are in use.
+    pub used_data: DataBlocks,
+    /// The total number of data blocks available to the thinpool.
+    pub total_data: DataBlocks,
+}
+
+#[derive(Debug, Clone, Copy)]
+/// Top-level thinpool status that indicates if it is working or failed.
+pub enum ThinPoolStatus {
+    /// The thinpool is working.
+    Good((ThinPoolWorkingStatus, ThinPoolBlockUsage)),
+    /// The thinpool is in a failed condition.
+    Fail,
+}
+
+#[derive(Debug, Clone, Copy)]
+/// Indicates if a working thinpool is working optimally, or is
+/// experiencing a non-fatal error condition.
+pub enum ThinPoolWorkingStatus {
+    /// The pool is working normally.
+    Good,
+    /// The pool has been forced to transition to read-only mode.
+    ReadOnly,
+    /// The pool is out of space.
+    OutOfSpace,
+    /// The pool needs checking.
+    NeedsCheck,
+}
 
 /// Use DM to create a "thin-pool".  A "thin-pool" is shared space for
 /// other thin provisioned devices to use.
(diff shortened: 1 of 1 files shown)


src/thinpooldev.rs
```diff
@@ -99,6 +136,65 @@ impl ThinPoolDev {
...
+    pub fn status(&self, dm: &DM) -> DmResult<ThinPoolStatus> {
+        let (_, mut status) =
+            try!(dm.table_status(&DevId::Name(self.dev_info.name()), DmFlags::empty()));
+
+        if status.len() != 1 {
+            return Err(DmError::Io(io::Error::new(io::ErrorKind::InvalidData,
+                                                  "Expected 1 line from thin pool status")));
+        }
+
+        let status_line = status.pop().unwrap().3;
```
mulkieran: This is a fine place for an ```expect()``` (instead of ```unwrap()```).

mulkieran: It might be more helpful if TargetLine had named fields...

mulkieran: Why don't LinearDev, etc. use TargetLine for their TargetLine tuples?

src/thinpooldev.rs
```diff
@@ -99,6 +136,65 @@ impl ThinPoolDev {
...
+            return Err(DmError::Io(io::Error::new(io::ErrorKind::InvalidData,
+                                                  "Kernel returned too few values from thin \
+                                                   pool status")));
+        }
+
+        let usage = {
+            let meta_vals = status_vals[1].split('/').collect::<Vec<_>>();
+            let data_vals = status_vals[2].split('/').collect::<Vec<_>>();
+            ThinPoolBlockUsage {
+                used_meta: meta_vals[0].parse::<u64>().unwrap(),
```
mulkieran: More of ```expect()``` here desirable.

src/thinpooldev.rs
```diff
@@ -5,6 +5,7 @@
 use std::fmt;
 use std::path::Path;
 use std::path::PathBuf;
+use std::io;
```
mulkieran: This could be considered ever so slightly out of order.

src/thinpooldev.rs
agrover: expect(): the line right before returns an error if len != 1 but ok, why not

I don't think fields in a tuple can have names. We'd have to make it a struct, which is a bigger undertaking...

They do now.

src/thinpooldev.rs
```diff
@@ -2,6 +2,7 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+use std::io;
```
mulkieran: "i" comes after "f"

src/thinpooldev.rs
```diff
@@ -103,6 +140,73 @@ impl ThinPoolDev {
...
+    pub fn status(&self, dm: &DM) -> DmResult<ThinPoolStatus> {
+        let (_, mut status) = try!(dm.table_status(&DevId::Name(self.dev_info.name()),
+                                                   DmFlags::empty()));
+
+        if status.len() != 1 {
+            return Err(DmError::Io(io::Error::new(io::ErrorKind::InvalidData,
+                                                  "Expected 1 line from thin pool status")));
+        }
+
+        let status_line = status.pop().expect("status didn't have one line").3;
```
mulkieran: I'ld phrase these things positively, because of the meaning of the word "expect". If a surprise() function existed, it would make more sense to phrase these negatively. So, I'ld go with something like
```expect("status has exactly one line")```

File: src/thinpooldev.rs [hunk 6]
```diff
@@ -103,6 +140,73 @@ impl ThinPoolDev {
...
+            return Err(DmError::Io(io::Error::new(io::ErrorKind::InvalidData,
+                                                  "Kernel returned too few values from thin \
+                                                   pool status")));
+        }
+
+        let usage = {
+            let meta_vals = status_vals[1].split('/').collect::<Vec<_>>();
+            let data_vals = status_vals[2].split('/').collect::<Vec<_>>();
+            ThinPoolBlockUsage {
+                used_meta: meta_vals[0]
```
mulkieran: Same positive change here and immediately following, as ```expect("status result has expected format")```

src/thinpooldev.rs
```diff
@@ -25,6 +26,42 @@ impl fmt::Debug for ThinPoolDev {
...
 }
 
+#[derive(Debug, Clone, Copy)]
+/// Contains values indicating the thinpool's used vs total
+/// allocations for metadata and data blocks.
+pub struct ThinPoolBlockUsage {
+    /// The number of metadata blocks that are in use.
+    pub used_meta: u64,
+    /// The total number of metadata blocks available to the thinpool.
+    pub total_meta: u64,
```
trgill: Why don't we use DataBlocks type for used_meta and total_meta?

agrover: The question I have is should we define a new `MetaDataBlocks` type or reuse `DataBlocks`. What do you think?

src/thinpooldev.rs
agrover: Alphabetizing use declarations isn't in the style guide...

agrover: Oops! Actually it is, but it's not required.

src/thinpooldev.rs
trgill: you thinking we could use a type to combine used_meta and total_meta?  that seems like a good idea to me.

src/thinpooldev.rs
```diff
@@ -103,6 +139,62 @@ impl ThinPoolDev {
...
+    /// Get the current status of the thinpool.
+    pub fn status(&self, dm: &DM) -> DmResult<ThinPoolStatus> {
+        let (_, mut status) = try!(dm.table_status(&DevId::Name(self.dev_info.name()),
+                                                   DmFlags::empty()));
+
+        assert_eq!(status.len(),
+                   1,
+                   "Kernel must return 1 line from thin pool status");
+
+        let status_line = status.pop().unwrap().3;
```
mulkieran: -> ```expect("assertion above holds")```

Right now, in stratisd, use of ```expect``` means we have thought this thing through and we have a good reason for it to always succeed. ```unwrap``` means that is not the case. This is very easy for grep to deal with, is helpful now, and will be nice if we end up thinking about the flight-recorder idea.

PR #22, opened 2017-04-17
Add ThinDev::status()

agrover:
Similar to ThinPoolDev::status(), this reads the thin device's status and
converts it from the kernel's text format into proper types.

### DIFF
--- a/src/thindev.rs
+++ b/src/thindev.rs
@@ -27,6 +27,16 @@ impl fmt::Debug for ThinDev {
     }
 }
 
+#[derive(Debug, Clone, Copy)]
+/// Thin device status.
+pub enum ThinStatus {
+    /// Thin device is good. Includes number of mapped sectors, and
+    /// highest mapped sector.
+    Good((Sectors, Sectors)),
+    /// Thin device is failed.
+    Fail,
+}
+
 /// support use of DM for thin provisioned devices over pools
 impl ThinDev {
     /// Use the given ThinPoolDev as backing space for a newly constructed
@@ -76,6 +86,31 @@ impl ThinDev {
             .devnode()
             .ok_or(DmError::Dm(InternalError("No path associated with dev_info".into())))
 
+    }
+
+    /// Get the current status of the thin device.
+    pub fn status(&self, dm: &DM) -> DmResult<ThinStatus> {
+        let (_, mut status) = try!(dm.table_status(&DevId::Name(&self.dev_info.name()),
+                                                   DmFlags::empty()));
+
+        assert!(status.len() == 1,
+                "Kernel must return 1 line from thin status");
+
+        let status_line = status.pop().expect("assertion above holds").3;
+        if status_line.starts_with("Fail") {
+            return Ok(ThinStatus::Fail);
+        }
+
+        let status_vals = status_line.split(' ').collect::<Vec<_>>();
+        assert!(status_vals.len() >= 2,
+                "Kernel must return at least 2 values from thin pool status");
+
+        Ok(ThinStatus::Good((Sectors(status_vals[0]
+                                         .parse::<u64>()
+                                         .expect("mapped sector count value must be valid")),
+                             Sectors(status_vals[1]
+                                         .parse::<u64>()
+                                         .expect("highest mapped sector value must be valid")))))
     }
 
     /// Remove the device from DM


src/thindev.rs
```diff
@@ -2,6 +2,7 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+use std::io;
```
mulkieran: swap lines.

src/thindev.rs
```diff
@@ -78,6 +88,28 @@ impl ThinDev {
...
+        let (_, mut status) = try!(dm.table_status(&DevId::Name(&self.dev_info.name()),
+                                                   DmFlags::empty()));
+
+        if status.len() != 1 {
+            return Err(DmError::Io(io::Error::new(io::ErrorKind::InvalidData,
+                                                  "Expected 1 line from thin status")));
+        }
+
+        // We should either get 1 line or the kernel is broken
+        let status_line = status.pop().expect("status didn't have one line").3;
```
mulkieran: Go for positive, here as "status has exactly one line".

src/thindev.rs
```diff
@@ -78,6 +88,28 @@ impl ThinDev {
...
+        if status_line.starts_with("Fail") {
+            return Ok(ThinStatus::Fail);
+        }
+        let status_vals = status_line.split(' ').collect::<Vec<_>>();
+
+        Ok(ThinStatus::Good(Sectors(status_vals[0]
+                                        .parse::<u64>()
+                                        .expect("invalid sectors from thin status"))))
```
mulkieran: Be positive! Perhaps "status_vals has expected format".

src/thindev.rs
```diff
@@ -78,6 +88,28 @@ impl ThinDev {
...
+                                                  "Expected 1 line from thin status")));
+        }
+
+        // We should either get 1 line or the kernel is broken
+        let status_line = status.pop().expect("status didn't have one line").3;
+        if status_line.starts_with("Fail") {
+            return Ok(ThinStatus::Fail);
+        }
+        let status_vals = status_line.split(' ').collect::<Vec<_>>();
+
```
mulkieran: Properly speaking, you should return an error if ```status_vals``` has length 0 as in PR #20.

src/thindev.rs
```diff
@@ -78,6 +88,28 @@ impl ThinDev {
...
+    pub fn status(&self, dm: &DM) -> DmResult<ThinStatus> {
+        let (_, mut status) = try!(dm.table_status(&DevId::Name(&self.dev_info.name()),
+                                                   DmFlags::empty()));
+
+        if status.len() != 1 {
+            return Err(DmError::Io(io::Error::new(io::ErrorKind::InvalidData,
+                                                  "Expected 1 line from thin status")));
+        }
+
+        // We should either get 1 line or the kernel is broken
```
trgill: Just checking that I understand this comment.... There is only one line in the DM table because ThinPoolDev provides what appears as continuous sectors to the ThinDev.  There would only be multiple lines in the dmtable when space is segmented.  That wouldn't make sense for a thin provisioned device.  Right?

agrover: yes, exactly.

src/thindev.rs
agrover: The expect message is if the unwrap fails, I believe.

agrover: ah I see what you're sayingnow , and it makes sense, but it's not how it's being used elsewhere in the crate, or in the main Rust API docs. For now I think it is best to stick to common usage, even if it is perverse.

src/thindev.rs
```diff
@@ -78,6 +88,31 @@ impl ThinDev {
...
 
+    /// Get the current status of the thin device.
+    pub fn status(&self, dm: &DM) -> DmResult<ThinStatus> {
+        let (_, mut status) = try!(dm.table_status(&DevId::Name(&self.dev_info.name()),
+                                                   DmFlags::empty()));
+
+        assert!(status.len() == 1,
+                "Kernel must return 1 line from thin status");
+
+        let status_line = status.pop().unwrap().3;
```
mulkieran: Same comment as for #20, or maybe just ```expect("status.len() == 1")```.

PR #23, opened 2017-04-19
ThinDev::new should take a reference to the ThinPoolDev

trgill:
ThinDev::new should take a reference to the ThinPoolDev
Change Segment::new() to take Sectors rather than u64


agrover:
👍

PR #31, opened 2017-04-24
Add ThinPoolDev::setup

agrover:
This function re-creates a previously-existing thinpool. The primary
difference between this and ThinPoolDev::new is that it calls thin_check
to verify metadata is in a good state before proceeding.


agrover:
This can go in after the reorg PR, so just review-prelim for the moment.

Issue #32, opened 2017-04-24
Re-export functions or structs?


agrover:
Now that things are a little better broken up by #28, I'd like to raise the question of which identifiers we should re-export at the top level. See [this](https://doc.rust-lang.org/book/crates-and-modules.html#re-exporting-with-pub-use).

I was thinking we make it nicer to use the HL API by re-exporting, but not re-export the LL stuff. @trgill what's your take?

trgill:
Just to make sure I understand. :-) 

You are thinking we hide the contents/functions of dm.rs, device.rs and deviceinfo.rs?

For example, we'd remove the dm parameter from ThinDev::new() etc?

agrover:
It's been a while since I had my head down in it, but I was just thinking we want to make it easy as we can to use the high level (HL) API, and still make the LL API available but tuck it down another level, perhaps. But, there are LL structs e.g. Device, DeviceInfo, DM that HL also needs, and these probably need to be at the toplevel too.

Do you think this goal of saving HL users from dealing with the LL details achievable? Or maybe it's just not realistic or that big a win, I don't know.

Specifically for `DM`.. passing it in makes the HL API more complicated but haven't yet seen a way, if it were not passed in, of avoiding pathological cases where `DM` structs get created/destroyed in huge, inefficient numbers.

trgill:
Yes, I think the goal of saving HL users from dealing with the details is achievable.  I think we are fairly close.  I'll put a PR together.

trgill:
@agrover the latest PRs address these issues, right?

agrover:
Yes I think so! Closing. Thanks!

Issue #25, opened 2017-04-21
Reorganize implementation for new, larger scope


agrover:
Now that devicemapper has a higher-level API in addition to its lower-level one, we may want to reorganize -- move the lower-level stuff out of lib.rs into its own file, re-export some `use` decls at the toplevel for convenience, such as `Segment`, `Sectors` and `Bytes`, `ThinDev`, and `ThinPoolDev`.

Also, changing LL functions to return `DmResult` instead of `io::Result`, and fixing any other things that no longer make sense.

PR #35, opened 2017-04-25
Update DM::* functions that return possible errors to use DmResult.  Make bin/test a unit test in dm.rs

trgill:
Update Lower Level API functions to return DmResult.
Remove pub from all most mods (except consts) and explicitly export high level API in lib.rs
Move bin/dmtest.rs to a unit test in dm.rs 
Corrected a couple spelling typos and added some comments


trgill:
@agrover please review.

My intent of this PR is to finish #25 and mostly address #32.

src/dm.rs
```diff
@@ -324,7 +324,7 @@ impl DM {
         };
 
         if new_name.as_bytes().len() > max_len {
-            return Err(Error::new(Other, format!("New name {} too long", new_name)));
+            return Err(DmError::Io(Error::new(Other, format!("New name {} too long", new_name))));
```
agrover: The old code didn't define its own error type and so kinda crammed everything into an io error. This is one place where maybe we can make a better choice now. How about making this an InternalError?

src/dm.rs
```diff
@@ -744,3 +744,41 @@ impl DM {
...
+    #[test]
+    fn test_basics() {
+        let dmi = DM::new().unwrap();
+
+        println!("Calling version()");
+        let x = dmi.version().unwrap();
+        println!("{:?}", x);
+
+        println!("Calling list_devices()");
+        let x = dmi.list_devices().unwrap();
```
agrover: This still needs more work but that can happen in a separate PR. This is still a big step in the right direction.

src/dm.rs
```diff
@@ -30,6 +30,7 @@ use types::TargetLine;
 use util::slice_to_null;
 
 /// Major numbers used by DM.
+#[allow(dead_code)]
```
agrover: yeah I think it can go. We can always resurrect it.

trgill: Ok

src/dm.rs
trgill: Ok

agrover:
👍

PR #37, opened 2017-05-01
Implement Step for Sectors and Bytes

mulkieran:
1) It does no harm, since Step is mathematically well-establishsed.
2) ranged_set requires that the things in the ranges impl Step. Ranges of
Sector offsets do kind of make sense, so there is a motivation.


agrover:
Code looks good, but the case for applying this depends on using `ranged_set` long term. 

After looking at that code to assess the amount of work to better handle ranges, I think it might be easier to implement what we need ourselves. Prior to finding ranged_set, I had tried implementing what RangeAllocator needed directly by wrapping a BTreeMap<offset, length>. This got ugly when it came to do lookups/merges, but the `range` method newly stabilized in Rust 1.17 could simplify this a lot.

mulkieran:
I realize that nothing uses step currently but ```ranged_set```. On the other hand, it is a small bit of code with sound mathematical backing, so it is unlikely to do much harm.

PR #39, opened 2017-05-11
Updates for rustfmt 0.8.3

trgill:
Update the code using rustfmt 0.8.3 as needed
Update the Makefile to include a format check as part of CI
Allow for 110 character line length.  The default is 100.  The code had a pointer to a long
URL.  I think allowing 110 is better than removing the comment.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,10 @@
-check: build test
+check: fmt build test
+
+${HOME}/.cargo/bin/cargo-fmt:
+	cargo install rustfmt --vers 0.8.3
+
+fmt: ${HOME}/.cargo/bin/cargo-fmt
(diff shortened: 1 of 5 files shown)


trgill:
@agrover @mulkieran review please.  I think mostly trivial changes.

I added a rustfmt.tom to allow 110 chars.  There is a comment is dm.rs that points to this URL:

https://git.kernel.org/cgit/linux/kernel/git/torvalds/linux.git/tree/Documentation/device-mapper

After indentation it ends up to be around 105 chars.

mulkieran:
Better to split the URL and accept the defaults, I would think. I wouldn't like the lines to get any longer.

trgill:
Ok, I removed the rustfmt.toml and split the URL.

agrover:
👍

PR #40, opened 2017-05-11
Thindev setup & destroy

agrover:
Please especially look at the second commit -- it needs a `&ThinPoolDev` to send the message, but it's not clear how to acquire this in the likely caller, `StratFilesystem::destroy()`.

This suggests we should consider refactoring something on either the dm-rs or stratisd sides to make this work better. Thoughts? Maybe we don't need destroy in the Filesystem trait, just have Pool::destroy_filesystems do it directly?

### DIFF
--- a/src/thindev.rs
+++ b/src/thindev.rs
@@ -41,8 +41,8 @@ pub enum ThinStatus {
 
 /// support use of DM for thin provisioned devices over pools
 impl ThinDev {
-    /// Use the given ThinPoolDev as backing space for a newly constructed
-    /// thin provisioned ThinDev returned by new().
+    /// Use the given ThinPoolDev as backing space for a newly
+    /// constructed thin provisioned ThinDev.
     pub fn new(name: &str,
                dm: &DM,
                thin_pool: &ThinPoolDev,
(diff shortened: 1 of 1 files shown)


src/thindev.rs
```diff
@@ -125,7 +125,16 @@ impl ThinDev {
                                          .expect("highest mapped sector value must be valid")))))
     }
 
-    /// Remove the device from DM
+    /// Remove the thin device.
+    pub fn destroy(self, dm: &DM, thin_pool: &ThinPoolDev) -> DmResult<()> {
+        let id = self.thin_id;
+        try!(self.teardown(dm));
+        try!(thin_pool.message(dm, &format!("delete {}", id)));
```
trgill: How frequently will the thin_pool.message() fail?  Will the thin_pool at some point realize the resource should be released if the message() does fail?

agrover: From looking at the kernel code, it will fail only if the message doesn't have the proper number of arguments (1) or if the argument is not the thin_id of a valid thin device. 

There's no way to enumerate which ids are valid, so it's really up to the user (us) to keep track, and no way for thinpool to detect when data is unused and free it on its own.

src/thindev.rs
```diff
@@ -51,16 +51,26 @@ impl ThinDev {
                -> DmResult<ThinDev> {
 
         try!(thin_pool.message(dm, &format!("create_thin {}", thin_id)));
+        ThinDev::setup(name, dm, thin_pool, thin_id, length)
+    }
+
+    /// Create the device for an existing thin volume.
+    pub fn setup(name: &str,
```
trgill: I think the pub is unnecessary. (?)

agrover: I think this needs to be part of the public API because callers will want to be able to recreate a thindev based upon existing data in the thinpool, such as after a reboot. They can't call `new` because `new` sends the "create_thin" message, which will fail if the given thin_id already exists.

trgill: Ah, right - I see.

src/thindev.rs
```diff
@@ -51,16 +51,26 @@ impl ThinDev {
...
+                 dm: &DM,
+                 thin_pool: &ThinPoolDev,
+                 thin_id: u32,
+                 length: Sectors)
+                 -> DmResult<ThinDev> {
         try!(dm.device_create(name, None, DmFlags::empty()));
         let id = &DevId::Name(name);
```
mulkieran: I happened to notice that it looks a little wierd that ```id``` is a reference to a DevId, but then a reference to ```id``` is passed as an argument.

PR #43, opened 2017-05-12
Add ThinDev::extend()

agrover:
This method extends the thin volume's logical size by the number of
sectors given.

Since this involves reloading the device's table after initialization, it
was necessary to remember the associated thinpool's dstr so we could
properly reload its table. Modifying ThinDev::dm_table args was also
needed.

### DIFF
--- a/src/lineardev.rs
+++ b/src/lineardev.rs
@@ -6,7 +6,7 @@ use std::fmt;
 use std::fs::File;
 use std::path::PathBuf;
 
-use consts::DmFlags;
+use consts::{DmFlags, DM_SUSPEND};
 use deviceinfo::DeviceInfo;
 use dm::{DM, DevId};
 use result::{DmResult, DmError, InternalError};
@@ -18,6 +18,7 @@ use util::blkdev_size;
 pub struct LinearDev {
     /// Data about the device
     pub dev_info: DeviceInfo,
+    segments: Vec<Segment>,
 }
 
 impl fmt::Debug for LinearDev {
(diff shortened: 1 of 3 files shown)


agrover:
@trgill Added a second commit, could you take a look at that one too? (Upon reflection, I probably should've just merged the first commit you approved and created another PR for the second.....)

src/lineardev.rs
```diff
@@ -32,21 +33,27 @@ impl LinearDev {
...
-        Ok(LinearDev { dev_info: dev_info })
+        Ok(LinearDev {
+               dev_info: dev_info,
+               segments: block_devs,
+           })
     }
 
     /// Generate a Vec<> to be passed to DM.  The format of the Vec entries is:
     /// <logical start sec> <length> "linear" /dev/xxx <start offset>
     fn dm_table(block_devs: &[Segment]) -> Vec<TargetLine> {
```
trgill: I should have used the variable name segments instead of block_devs, or would you prefer segs?

I can open a PR to fix.

src/thindev.rs
```diff
@@ -120,6 +121,22 @@ impl ThinDev {
                                          .expect("highest mapped sector value must be valid")))))
     }
 
+    /// Extend the thin volume's (virtual) size by the number of
```
mulkieran: Should it really be a "volume" or is it still a "device"?

src/lineardev.rs
agrover: sure, sounds good. I have no preference on the new name.

Could you please also add that little `segments` method that we need, while you're doing a new PR? 😄

src/lineardev.rs
```diff
@@ -32,21 +33,27 @@ impl LinearDev {
     /// Construct a new block device by concatenating the given block_devs
     /// into linear space.  Use DM to reserve enough space for the stratis
     /// metadata on each DmDev.
-    pub fn new(name: &str, dm: &DM, block_devs: &[Segment]) -> DmResult<LinearDev> {
+    pub fn new(name: &str, dm: &DM, block_devs: Vec<Segment>) -> DmResult<LinearDev> {
```
mulkieran: The variable naming _is_ a little wierd here.

src/lineardev.rs
```diff
@@ -63,6 +70,42 @@ impl LinearDev {
         table
     }
 
+    /// Extend an existing LinearDev with additional new segments.
```
mulkieran: Why are new segments certain to be ordered?

mulkieran: Why are new segments certain not to be adjacent?

src/lineardev.rs
agrover: will be fixed by @trgill 's separate PR.

src/lineardev.rs
agrover: > Why are new segments certain to be ordered?

They have an order just by virtue of being in a `Vec`, and that's the order the new, extended LinearDev will use.

> Why are new segments certain not to be adjacent?

Hmm, I guess there's nothing to prevent this right now. I'll open an issue.

src/thindev.rs
agrover: ok, "device" is clearer.

agrover:
* Fixed `ThinDev::extend()` documentation to refer to device rather than volume.

Other stuff I'd like to do via separate PRs. Opened #44 to track. @mulkieran sound ok?

src/lineardev.rs
mulkieran: They have the ordering of being in a Vec, i.e., there's a 0th, and a 1st, etc. But the coalescing doesn't work unless the ordering in the Vec is also some kind of logical ordering of the segment elements. In other words, why is it only the first in the new one that might be contiguous with only the last in the previous one?

agrover: > why is it only the first in the new one that might be contiguous with only the last in the previous one?

It's only the last (of the existing segments) because one can't change the offsets of where existing segments show up in the mapping. Extending an existing segment other than the last changes the offsets of all later data.

It's only the first (of the new segments) for much the same reason: There may already be data on the new segments and we can't reorder them safely. The only spot where we can combine two segments into one without reordering is last-existing to first-new.

mulkieran:
Please go ahead...btw, Add would be a bad name for a Segments extend operation, it's not commutative.

agrover:
thanks!

PR #48, opened 2017-05-18
Avoid redundannt methods

mulkieran:
Sectors and Bytes are differently interconvertible by means of the
methods bytes() and sectors(). sectors() has been carefully defined
to do something sensible when the value is not a whole number of sectors.

If thindev size will always be in granularity of sectors, then it is
reasonable for the unique size() method to return Sectors.

On the other hand, if thindev size will not always be in granularity of
sectors, then it would be best to have the unique size() method return
Bytes.

This commit assumes that the thindev size will always in granularity of
sectors.


agrover:
👍

PR #49, opened 2017-05-18
Make Segment implement Serialize and Deserialize

mulkieran:
Stratis would like to store and retrieve information encoded
as Segments. Do the same for Device, since Segment uses it.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,5 +18,6 @@ clippy = { version = "0", optional = true }
 newtype_derive = "0.1"
 custom_derive = "0.1"
 serde = "1"
+serde_derive = "1"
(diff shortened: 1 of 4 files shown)


mulkieran:
This is experiencing a somewhat puzzling build error on my machine.

mulkieran:
@agrover Do you have any idea what might be going on? syn has a lot of dependent crates. This feels like maybe it qualifies as a cargo bug.

agrover:
you branch builds for me. Try also `git clean -fxd` maybe?

mulkieran:
Yes. I think the key is to remove Cargo.lock, which cargo clean doesn't do.

PR #51, opened 2017-05-18
Adapt for change in 4.12+ kernels

### LINKED ISSUE #50: slice index starts at 312 but ends at 305
@trgill is seeing this and I am too. I'm not sure when it started.
```
thread 'real_test_linear_device' panicked at 'slice index starts at 312 but ends at 305', src/libcore/slice.rs:584

10: <core::ops::Range<usize> as core::slice::SliceIndex<T>>::index
             at /builddir/build/BUILD/rustc-1.17.0-src/src/libcore/slice.rs:707
  11: core::slice::<impl core::ops::Index<I> for [T]>::index
             at /builddir/build/BUILD/rustc-1.17.0-src/src/libcore/slice.rs:560
  12: <collections::vec::Vec<T> as core::ops::Index<core::ops::Range<usize>>>::index
             at /builddir/build/BUILD/rustc-1.17.0-src/src/libcollections/vec.rs:1442
  13: devicemapper::dm::DM::do_ioctl
             at /home/agrover/git/stratis/devicemapper-rs/src/dm.rs:152
  14: devicemapper::dm::DM::device_create
             at /home/agrover/git/stratis/devicemapper-rs/src/dm.rs:253
  15: devicemapper::lineardev::LinearDev::new
             at /home/agrover/git/stratis/devicemapper-rs/src/lineardev.rs:37
  16: real_tests::util::simple_tests::test_linear_device
             at ./tests/util/simple_tests.rs:131
```

agrover:
See https://www.spinics.net/lists/stable/msg170390.html and 4617f564c in
the Linux kernel repo.

We were assuming hdr.data_size was always greater or equal to
hdr.data_start, but this is no longer the case.

This kernel change is also widely backported to previous kernel versions'
stable branches.

fixes #50

### DIFF
--- a/src/dm.rs
+++ b/src/dm.rs
@@ -145,7 +145,8 @@ impl DM {
         hdr_slc.clone_from_slice(&v[..hdr.data_start as usize]);
 
         // Maybe we got some add'l data back?
-        Ok(v[hdr.data_start as usize..hdr.data_size as usize].to_vec())
+        let new_data_off = cmp::max(hdr.data_start, hdr.data_size);
+        Ok(v[hdr.data_start as usize..new_data_off as usize].to_vec())
     }
 
     /// Devicemapper version information: Major, Minor, and patchlevel versions.


src/dm.rs
```diff
@@ -145,7 +145,8 @@ impl DM {
         hdr_slc.clone_from_slice(&v[..hdr.data_start as usize]);
 
         // Maybe we got some add'l data back?
-        Ok(v[hdr.data_start as usize..hdr.data_size as usize].to_vec())
+        let new_data_off = cmp::max(hdr.data_start, hdr.data_size);
```
trgill: Is this right?

If we are using new_data_off as the end point in an array buffer, I'd have expected it to be:

let new_data_off = hdr.data_start + hdr.data_size

agrover: Yes I think this is right. Despite its name, `data_size` is the offset from the start of the buffer to where the data ends, *not* a length. This actually worked out ok previously b/c it matches up with what we need for start and end index values to get the data. The API change means that for ioctls that return no data, the end offset (data_size) is now actually before the start offset (data_start) because the 7 byte padding is no longer copied back to the user, and data_size reflects this reduction.

PR #55, opened 2017-05-22
Remove some unneeded methods from Segment

mulkieran:
If a Segment's fields are going to be all public, these methods are not
appropriate.


agrover:
👍

PR #59, opened 2017-05-26
Add a default thin-pool block size.

trgill:
Used the suggested value for "people primarily interested in thin provisioning":

https://www.kernel.org/doc/Documentation/device-mapper/thin-provisioning.txt


mulkieran:
I'ld like to see the constant used by something before it goes in...I'm hoping for a follow up commit.

PR #60, opened 2017-05-30
Change return type argument to TargetLine

mulkieran:
It is already that type, now use the name TargetLine for that type.


mulkieran:
@trgill review?

PR #62, opened 2017-05-30
Add a checked_add() method for Bytes and Sectors

mulkieran:
Sometimes, we want to explicitly check if the values will overflow on
addition.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -51,6 +51,15 @@ macro_rules! unsigned_rem {
     }
 }
 
+macro_rules! checked_add {
+    ($T: ident) => {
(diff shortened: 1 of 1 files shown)


mulkieran:
checked multiplication and checked remainder are both trickier concepts, so I opted not to do them. ```checked_mul``` is tricky, because there is no associated trait, it must be defined for exactly one other argument type. ```checked_rem``` has the same problem, and I think, a few others having to do with the properties of remainders.

mulkieran:

mulkieran:
This is motivated by checks for overflow in https://github.com/stratis-storage/stratisd/pull/364, but that PR need not depend on this one.

PR #65, opened 2017-06-09
Since thindev ids are inconveniently special, give them a type

mulkieran:
The type can be amended or improved later, but at least it gives something
close to the maximally random number now.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -91,5 +91,5 @@ pub use lineardev::LinearDev;
 pub use result::{DmResult, DmError, ErrorEnum};
 pub use segment::Segment;
 pub use thinpooldev::{ThinPoolBlockUsage, ThinPoolDev, ThinPoolStatus, ThinPoolWorkingStatus};
-pub use thindev::{ThinDev, ThinStatus};
+pub use thindev::{ThinDev, ThinDevId, ThinStatus};
 pub use types::{Bytes, DataBlocks, Sectors};
--- a/src/thindev.rs
+++ b/src/thindev.rs
@@ -5,6 +5,8 @@
 use std::fmt;
 use std::path::PathBuf;
 
+use serde;
+
 use consts::{DmFlags, DM_SUSPEND};
 use deviceinfo::DeviceInfo;
 use dm::{DM, DevId};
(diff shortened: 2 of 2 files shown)


mulkieran:
@agrover @trgill review?

mulkieran:
It would be a convenience if this were merged and a new version were made today.

agrover:
We need to have a way to make a `ThinDevId` from a u32 for calling `setup`.

Also, I have doubts about the value of a way to generate random thin_ids in this crate, when the most important thing is that they not conflict with already used values, and the crate user is the one in position to ensure that.

mulkieran:
We can make ThinDevId serializable, which will allow us to use the type in our definition of the ThinPoolDevSave struct. That will allow us to make the values from u32s in the only place that I am aware of that we need to. I've done that part. We can implement From if we have to.

We can implement Hash and Eq, so that the client can, w/out difficulty, compare and make sure that no duplicates, etc. I've done that, also.

mulkieran:
The test, however, is completely wrong and should be fixed.

mulkieran:
rebase.

mulkieran:
It is possible that what devicemapper really needs is a trait, which stratisd can implement...

agrover:
Like I said in #67, I like this approach better but we just need a way for the client to be able to make a ThinDevId from a given number, rather than using rand.

mulkieran:
@agrover please take a look. I started to think that simply returning a result for new method made more sense, so I gave that a try.

src/thindev.rs
```diff
@@ -14,10 +16,59 @@ use types::TargetLine;
 
 use types::Sectors;
 
+const THIN_DEV_ID_LIMIT: u64 = 16777216; // 2 ^ 24
```
agrover: better to express in hex? 0xFFFFFF? Or, 0x1000000 I guess, if it's being used as the first not-possible value.

src/lib.rs
```diff
@@ -60,6 +60,8 @@ extern crate bitflags;
 #[macro_use]
 extern crate log;
 
+extern crate rand;
```
agrover: no longer needed

src/thindev.rs
```diff
@@ -14,10 +16,59 @@ use types::TargetLine;
...
+
+#[derive(Debug, Clone, Copy, Eq, PartialEq, Hash)]
+/// A thindev id is a 24 bit number, i.e., not a power of 2.
```
agrover: maybe say "bit width not a power of 2"

src/thindev.rs
```diff
@@ -14,10 +16,59 @@ use types::TargetLine;
...
+pub struct ThinDevId {
+    value: u32,
+}
+
+impl ThinDevId {
+    /// Make a new ThinDevId.
+    /// Return an error if value is too large to represent in 24 bits.
+    pub fn new_u64(value: u64) -> DmResult<ThinDevId> {
```
agrover: super-nitpick: not just `new`? `_u64` seems redundant

agrover: DmResult, ok.

mulkieran: I'ld rather leave things open to new from other categories of values, hence the ```_u64```.

src/lib.rs
mulkieran: Whoops!

PR #66, opened 2017-06-09
Add a Makefile target that runs only non-root tests

mulkieran:
It is convenient to run basic unit tests and non-running doc tests
on one's own system.

On the CI, continue to run all tests as root.

This is a somewhat temporary expedient. If more tests that require root
are added, it may pay to better distinguish between the different types of
tests.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-check: fmt build test
+check: fmt build sudo_test
 
 ${HOME}/.cargo/bin/cargo-fmt:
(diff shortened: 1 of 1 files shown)


mulkieran:
Cherry-picked from #65.

@trgill review? AFAIU, there was no objection to this commit and it is not really related to the other one in #65, so it might as well go in now.

agrover:
👍

PR #67, opened 2017-06-12
Make ThinDev implementation generic on its id type

mulkieran:
Clients can implement the id, but it has a type that can be recognized.
This is an improvement over the uninformative type u32.


mulkieran:
This is an alternative to #65 .

mulkieran:
@agrover @trgill review?

agrover:
Do we need a trait, or can we just define a "u24" type `ThinDevId`, with `new` and `new_unchecked` methods along the lines of https://docs.rs/sample/0.6.2/sample/types/u24/struct.U24.html?

agrover:
Heh, I should've looked at #65!!!

mulkieran:
We may be going with #65.

mulkieran:
Superceded by #65.

Issue #52, opened 2017-05-19
I'm not certain that the meta_ and data_ devices in a thin pool are certain to be linear devs


mulkieran:

mulkieran:
I would think that there would be quite a few different kinds of devices that might occupy that role.

agrover:
Agree. There's probably space for some (trait-based?) generalization here at some point.

mulkieran:
The need for trait-based generalization is getting more urgent, see #70 for examples.

agrover:
Could be a fun little project.

mulkieran:
Except for the "fun" and the "little".

mulkieran:
Shared:

* ```name()```
* ```dstr()```
* ```size()```
* ```devnode()```
* ```teardown()```

Is that useful?

agrover:
After further reflection, I think this may be a theoretical library restriction in `ThinPoolDev` but not an issue that Stratis will actually run into. Even in 2.0, Stratis's Flex layer is LinearDevs to handle spanning & extending across multiple lower-level devices if necessary. The LinearDevs will just be on top of RaidDevs and/or IntegrityDevs instead of base blockdevs.

Even ignoring Stratis, the value of a ThinPoolDev on top of something other than a LinearDev may not be great. You could presumably put a ThinPoolDev on top of two RaidDevs, but they each cannot be extended (without inserting a LinearDev! heh) and probably are the wrong size ratio of meta to data device.

mulkieran:
Might as well leave as LinearDev...at least for now. We don't really know how to extend all DmDevices, we might need a subcategory.

PR #68, opened 2017-06-12
Add extend to thinpooldev

trgill:
extend() function for ThinPoolDev is needed to allow users to extend the segments allocated
to the meta and data backing devices.

### DIFF
--- a/src/dm.rs
+++ b/src/dm.rs
@@ -487,6 +487,21 @@ impl DM {
         Ok(DeviceInfo { hdr: hdr })
     }
 
+    /// Reload the table for a device
+    pub fn table_reload<T1, T2>(&self,
+                                dm: &DM,
+                                id: &DevId,
+                                table: &[(u64, u64, T1, T2)])
+                                -> DmResult<()>
+        where T1: Borrow<str>,
+              T2: Borrow<str>
+    {
+        try!(dm.table_load(id, table));
+        try!(dm.device_suspend(id, DM_SUSPEND));
+        try!(dm.device_suspend(id, DmFlags::empty()));
+        Ok(())
+    }
+
     /// Clear the "inactive" table for a device.
     pub fn table_clear(&self, name: &DevId) -> DmResult<DeviceInfo> {
         let mut hdr: dmi::Struct_dm_ioctl = Default::default();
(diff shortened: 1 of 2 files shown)


agrover:
This currently extends the underlying meta or data devs, but from looking at what Froyo does it is also likely necessary to reload the thinpool dev for the new resources to get picked up. Especially a larger data dev, since the length of the thinpool must match the length of the underlying data dev, this must be updated.

trgill:
Added test code in https://github.com/stratis-storage/stratisd/pull/417.

agrover:
This is good for now. 👍 

Although I'd like to tag DM::table_reload (and DM::wait_for_dm) as candidates that may ultimately not belong at the lowest level, struct DM, but in some future midlevel behavior-common-to-all-targets thing, as mentioned in #52 .

src/dm.rs
```diff
@@ -487,6 +487,21 @@ impl DM {
         Ok(DeviceInfo { hdr: hdr })
     }
 
+    /// Reload the table for a device
+    pub fn table_reload<T1, T2>(&self,
+                                dm: &DM,
+                                id: &DevId,
+                                table: &[(u64, u64, T1, T2)])
+                                -> DmResult<()>
+        where T1: Borrow<str>,
```
mulkieran: The constraints here should really be ```AsRef``` rather than ```Borrow``` as ```Borrow``` adds restrictions about hashing that AFAIK, are not helpful. Changing that should be simple and can be deferred until a later commit.

src/thinpooldev.rs
```diff
@@ -240,6 +243,29 @@ impl ThinPoolDev {
...
+                             &ThinPoolDev::dm_table(try!(self.data_dev.size()),
+                                                    self.data_block_size,
+                                                    self.low_water_mark,
+                                                    &self.meta_dev,
+                                                    &self.data_dev)));
+        Ok(())
+    }
+
+    /// Extend an existing meta device with additional new segments.
+    pub fn extend_meta(&mut self, dm: &DM, new_segs: Vec<Segment>) -> DmResult<()> {
```
mulkieran: I don't think these two methods ought to be consuming ```new_segs```...I feel the same about ```LinearDev::extend```.

PR #74, opened 2017-06-19
Changes to devnode methods

mulkieran:


mulkieran:
@agrover @trgill review?

agrover:
Are there instances in our code where a caller might treat an `Err` and `Ok(None)` differently? If not, then maybe we should just make life easier on the caller and just return an `Err` if the devnode isn't found. Especially since this seems to cause other APIs to now need to return Option when they didn't before.

mulkieran:
Added a slight reformatting, as rustfmt was doing some funky indenting in two places. No other changes. rebased.

mulkieran:
I think it is the right thing to do. Currently, in stratisd, it won't be useful to distinguish between None and Err. But, I think as some of the code sophisticates, it will. But we can wait a bit, to give cooler heads a chance to prevail.

agrover:
Jeff Moyer says if you create a device, it won't show up in /proc/partitions until after `/dev/<whatever>` exists. This is because the kernel itself creates these in devtmpfs (/dev) as opposed to a udev rule doing it.

So @mulkieran I'd like to know more about the device-without-devnode case you've seen, and if we can understand it to draw a bright line between that, and the devices we're interested in, listed in /proc/partitions, perhaps?

agrover:
Even better!

agk just stated the devnode for a device is always /dev/dm-X where x is the minor number. We should be able to just return a `PathBuf` of `format!("/dev/dm-{}", self.minor)`

mulkieran:
What I am quite confident of is that there are devices that do not have devnodes, _ever_. This is why pyudev, and other libraries that follow its example, may return None for their ```devnode()``` method on a device.

mulkieran:
I must have missed some of this discussion. But if it means there's a way to do no worse than by reading /proc/partitions without reading /proc/partitions, then it's a good thing.

agrover:
> What I am quite confident of is that there are devices that do not have devnodes, ever.

OK, but what I'm hearing from agk is that there are no *DM block devices* without devnodes, so it's possible to safely always return a devnode path, given a `devicemapper::Device`.

(And that the devnode is derivable from the minor number, so we don't actually have to parse /proc/partitions.)

PR #82, opened 2017-06-30
Improve CI

mulkieran:
Break out the various targets into separate runs and add clippy target
back in as an allowed failure.


mulkieran:
@trgill review?

PR #83, opened 2017-06-30
Extend real tests, those that require sudo

mulkieran:
This extends the existing DM test and add some related header docs and small code changes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ build:
 	RUSTFLAGS='-D warnings' cargo build
 
 test:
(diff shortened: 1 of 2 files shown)


mulkieran:
@agrover @trgill review?

trgill:
not sure it makes sense to split out the tests by "sudo required".  unless we change the permissions, won't all the tests that open /dev/xxx "read/write"  require sudo?

mulkieran:
Yes, they will. But, we now have five tests that do not require 'sudo', and we should plan to have more.

agrover:
👍 last reviewer, merging.

PR #89, opened 2017-07-05
Add ThinDev::destroy()

agrover:
This function tears down the DM device, but also messages thinpool to free
up resources associated with the thindev.

### DIFF
--- a/src/thindev.rs
+++ b/src/thindev.rs
@@ -206,7 +206,17 @@ impl ThinDev {
         Ok(())
     }
 
-    /// Remove the device from DM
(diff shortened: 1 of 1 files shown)


src/thindev.rs
```diff
@@ -206,6 +206,15 @@ impl ThinDev {
         Ok(())
     }
 
+    /// Remove the thin device.
```
mulkieran: This could use an enhanced header comment, to explain why there needs to be an extra ```destroy()``` method, i.e., why all this action can't be folded into ```teardown()```.

agrover:
Tried to improve the method comment, as suggested.

PR #91, opened 2017-07-06
Make device_rename idempotent

mulkieran:
The ioctl that it wraps returns an error if the names are the same,
but it is much better if we don't.


mulkieran:
@agrover @trgill review?

agrover:
👍 . merging as last reviewer.

PR #95, opened 2017-07-07
Revert "Make device_rename idempotent"

mulkieran:
This reverts commit b54dcd1caa353b5510a980034761425a11926aae.

It only works properly for a name, but "name" might be a UUID.

### DIFF
--- a/src/dm.rs
+++ b/src/dm.rs
@@ -278,6 +278,7 @@ impl DM {
     ///
     /// Valid flags: DM_UUID
     ///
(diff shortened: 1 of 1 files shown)


mulkieran:
We now have two known problems with ```device_rename``` and we need to fix them both properly. Given how short the method is, it doesn't make sense to fix them both separately. So, by reverting, we get to restart from a clean slate, and fix them both at once.

@agrover @trgill review?

PR #96, opened 2017-07-07
Use macros to implement serialization and deserialization

mulkieran:
Everything that can be macroized for Sectors, Bytes, DataBlocks, has
been macroized.

Note that at some point, if DataBlocks is rewritten, how it is serialized
may change and the current macro may not work any more.


mulkieran:
@trgill review?

agrover:
👍

PR #97, opened 2017-07-07
define a type for MetaBlocks to be used for tracking thin-meta use

trgill:
Use a MetaBlocks type for tracking the meta data device utilization.

dm.table_status() for a thin pool returns the meta device usage in blocks. The
block size for a meta is a constant 8 sectors.  The MetaBlocks type
wraps these sizes.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -91,4 +91,4 @@ pub use result::{DmResult, DmError, ErrorEnum};
 pub use segment::Segment;
 pub use thinpooldev::{ThinPoolBlockUsage, ThinPoolDev, ThinPoolStatus, ThinPoolWorkingStatus};
 pub use thindev::{ThinDev, ThinDevId, ThinStatus};
-pub use types::{Bytes, DataBlocks, Sectors};
+pub use types::{Bytes, DataBlocks, MetaBlocks, Sectors};
--- a/src/types.rs
+++ b/src/types.rs
@@ -11,6 +11,10 @@ use std::ops::{Div, Mul, Rem, Add};
 
 use serde;
 
+/// a kernel defined block size constant for a DM meta device
+/// defined in drivers/md/persistent-data/dm-space-map-metadata.h line 12
+const META_BLOCK_SIZE: Sectors = Sectors(8);
+
 // macros for implementing serialize and deserialize on all types
 macro_rules! serde {
     ($T: ident) => {
--- a/src/types.rs
+++ b/src/types.rs
@@ -113,6 +113,24 @@ custom_derive! {
     #[derive(NewtypeAdd, NewtypeAddAssign,
              NewtypeDeref,
              NewtypeFrom,
+             NewtypeSub,
+             Debug, Clone, Copy, Eq, PartialEq, PartialOrd, Ord)]
+    /// A type for Meta Data blocks as used by the thin pool.
+    /// MetaBlocks have a kernel defined constant size of META_BLOCK_SIZE
+    pub struct MetaBlocks(pub u64);
+}
+
+impl MetaBlocks {
+    /// Return the number of Sectors in the MetaBlocks.
+    pub fn sectors(self) -> Sectors {
+        self.0 * META_BLOCK_SIZE
+    }
+}
+
+custom_derive! {
+    #[derive(NewtypeAdd, NewtypeAddAssign,
+             NewtypeDeref,
+             NewtypeFrom,
              NewtypeSub, NewtypeSubAssign,
              Debug, Default, Clone, Copy, Eq, PartialEq, PartialOrd, Ord)]
     /// Structure to represent bytes
--- a/src/thinpooldev.rs
+++ b/src/thinpooldev.rs
@@ -12,7 +12,7 @@ use dm::{DM, DevId};
 use lineardev::LinearDev;
 use result::{DmResult, DmError, ErrorEnum};
 use segment::Segment;
-use types::{DataBlocks, Sectors, TargetLine};
+use types::{DataBlocks, MetaBlocks, Sectors, TargetLine};
 
 /// DM construct to contain thin provisioned devices
 pub struct ThinPoolDev {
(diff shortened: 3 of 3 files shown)


mulkieran:
Why not incorporate this into a large PR where you use the const to define ```META_LOWATER```. Then it might turn out that it belongs in a different file entirely, or something.

mulkieran:
Alternatively, if it is only going to be used by Stratis, maybe it is a constant that should be defined in Stratis?

trgill:
Any client that monitors thin devices will need to know the block size of the meta device, right?

Why define it in Stratis?  It is a device mapper attribute.

mulkieran:
What happens if Stratis decides to use 16 Sectors instead, i.e., ignores this devicemapper constant entirely?

Whence the value of 8 Sectors?

trgill:
Stratis can only change the block size for data devices.

Meta device block size is not user controllable.  It is defined in the DM C code as:

#define DM_SM_METADATA_BLOCK_SIZE (4096 >> SECTOR_SHIFT)

I can see a concern about making a copy of a constant that could change in the future.  @agrover do you see a better approach?

agrover:
Yeah it's a kernel constant, it's not going to change under us without notice.

What were we going to use this for, again? To ensure when we expanded the metadata device it was a multiple of this?

trgill:
Currently in Stratis we've got:

const META_LOWATER: u64 = 512;

I'm trying to make it clear that we are talking about 512 blocks and each block is 8 sectors.   I suppose I could just add a comment if you'd prefer that.

mulkieran:
It looks to me like you need a MetaBlocks type that uses that constant value, somewhat analogous to DataBlocks, but less general because we know exactly the number of Sectors a MetaBlock has.

That would then be the type of ```ThinPoolBlockUsage::used_meta``` and ```ThinPoolBlockUsage::total_meta```?

trgill:
Ok

trgill:
@mulkieran @agrover I've made the updates you've requested.  Review please?

agrover:
👍 thanks!!!

src/types.rs
```diff
@@ -2,7 +2,7 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-use consts::SECTOR_SIZE;
+use consts::{META_BLOCK_SIZE, SECTOR_SIZE};
```
mulkieran: Might as well define ```META_BLOCK_SIZE``` here, rather than import it from consts. In a comment, you could state exactly where you got the value...source code file, name of constant, etc.

trgill: Ok, types.rs is not a "public" module, so I needed to add a line in lib.rs to export META_BLOCK_SIZE.

src/lib.rs
```diff
@@ -91,4 +91,4 @@ pub use result::{DmResult, DmError, ErrorEnum};
 pub use segment::Segment;
 pub use thinpooldev::{ThinPoolBlockUsage, ThinPoolDev, ThinPoolStatus, ThinPoolWorkingStatus};
 pub use thindev::{ThinDev, ThinDevId, ThinStatus};
-pub use types::{Bytes, DataBlocks, Sectors};
+pub use types::{Bytes, DataBlocks, MetaBlocks, META_BLOCK_SIZE, Sectors};
```
mulkieran: Let's not export ```META_BLOCK_SIZE``` until it is certain that we need to. It might be the case that we can do everything we need with the MetaBlocks type.

src/types.rs
```diff
@@ -11,6 +11,10 @@ use std::ops::{Div, Mul, Rem, Add};
 
 use serde;
 
+/// a kernel defined block size constant for a DM meta device
+/// defined in drivers/md/persistent-data/dm-space-map-metadata.h line 12
+pub const META_BLOCK_SIZE: Sectors = Sectors(8);
```
mulkieran: As part of not exporting the constant, this could be made private.

trgill: We need the constant in Stratis.  When we extend the meta device it needs to be a multiple of META_BLOCK_SIZE.

mulkieran: It is likely better to add functionality to MetaBlocks to accomplish the same thing.

trgill: Add a block_size() function to MetaBlocks?

trgill: Ok.  Done.

agrover:
after squashing, lgtm. 👍 

Time for another tag/release already?

src/types.rs
```diff
@@ -109,6 +113,24 @@ custom_derive! {
...
+             Debug, Clone, Copy, Eq, PartialEq, PartialOrd, Ord)]
+    /// A type for Meta Data blocks as used by the thin pool.
+    /// MetaBlocks have a kernel defined constant size of META_BLOCK_SIZE
+    pub struct MetaBlocks(pub u64);
+}
+
+impl MetaBlocks {
+    /// Return the number of Sectors in the MetaBlocks.
+    pub fn sectors(self) -> Sectors {
+        Sectors(self.0 * META_BLOCK_SIZE.0)
```
mulkieran: Just

```self.0 * META_BLOCK_SIZE```

ought to work.

File: src/types.rs [hunk 5]
```diff
@@ -109,6 +113,24 @@ custom_derive! {
...
+             Debug, Clone, Copy, Eq, PartialEq, PartialOrd, Ord)]
+    /// A type for Meta Data blocks as used by the thin pool.
+    /// MetaBlocks have a kernel defined constant size of META_BLOCK_SIZE
+    pub struct MetaBlocks(pub u64);
+}
+
+impl MetaBlocks {
+    /// Return the number of Sectors in the MetaBlocks.
+    pub fn sectors(self) -> Sectors {
+        Sectors(self.0 * META_BLOCK_SIZE)
```
mulkieran: Just

```self.0 * META_BLOCK_SIZE```...no ```Sectors(...```.

Feel free to merge once that is fixed.

PR #108, opened 2017-07-18
Implement multiplication and serde on all *Blocks types

mulkieran:
Leave out sum(), because it also requires implementing Default,
which might not be obvious for DataBlocks.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -109,6 +109,12 @@ custom_derive! {
 
 serde!(DataBlocks);
 
+unsigned_mul!(u64, DataBlocks);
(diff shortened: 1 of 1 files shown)


mulkieran:

trgill:
would it make sense to track the block size in the data blocks?  it would be nice to have a .sectors() function associated with DataBlocks.  PR looks good - don't hold for this question.

mulkieran:
It would make sense to track the block size in DataBlocks. But it is not clear how to do it because the block size is not a constant. There is an issue up about the problem.

agrover:
👍 , merging

PR #110, opened 2017-07-26
Refine new() and setup() methods

mulkieran:
This PR includes a bunch of changes.

* When making a new device still known to the kernel, it verifies that the given information matches the information known to the kernel.
* It removes the redundant size parameter from ```ThinPoolDev::new()``` and ```setup()```
* It makes running ```thin_check``` in ```ThinPoolDev::setup()``` conditional on the device being unknown to the kernel.
* It changes the return type of ```DM::parse_table_status()``` since the method prefers to panic in all cases where it might otherwise return an error.
* It trims trailing white space off the last entry in each line of the table returned by ```DM::parse_table_status()```.
* It add several tests to thinpooldev.rs.
* It does other small adjustements and fixes.

It addresses #120 but can not be certainly said to fix it, due to the uncertainty of what it means to be "in use".

### DIFF
--- a/src/thindev.rs
+++ b/src/thindev.rs
@@ -7,7 +7,7 @@ use std::path::PathBuf;
 
 use serde;
 
-use consts::DmFlags;
+use consts::{DM_STATUS_TABLE, DmFlags};
 use device::Device;
 use deviceinfo::DeviceInfo;
 use dm::{DM, DevId, DmName};
@@ -133,8 +133,6 @@ impl ThinDev {
     /// Set up an existing thindev.
     /// By "existing" is here meant that metadata for this thin device exists
     /// on the metadata device for its thin pool.
-    /// TODO: If the device is already known to the kernel, verify that kernel
-    /// model matches arguments.
     pub fn setup(name: &DmName,
                  dm: &DM,
                  thin_pool: &ThinPoolDev,
@@ -144,12 +142,15 @@ impl ThinDev {
 
         let id = DevId::Name(name);
         let thin_pool_device = thin_pool.device();
+        let table = ThinDev::dm_table(thin_pool_device, thin_id, length);
 
         let dev_info = if device_exists(dm, name)? {
-            // TODO: Verify that kernel's model matches arguments.
+            if dm.table_status(&id, DM_STATUS_TABLE)?.1 != table {
+                return Err(DmError::Dm(ErrorEnum::Invalid,
+                                       "specified data does not match kernel data".into()));
+            }
             Box::new(dm.device_status(&id)?)
         } else {
-            let table = ThinDev::dm_table(thin_pool_device, thin_id, length);
             Box::new(device_create(dm, name, &id, &table)?)
         };
 
--- a/src/lineardev.rs
+++ b/src/lineardev.rs
@@ -5,7 +5,7 @@
 use std::fmt;
 use std::path::PathBuf;
 
-use consts::DmFlags;
+use consts::{DM_STATUS_TABLE, DmFlags};
 use device::Device;
 use deviceinfo::DeviceInfo;
 use dm::{DM, DevId, DmName};
@@ -58,8 +58,6 @@ impl LinearDev {
     /// Warning: If the segments overlap, this method will succeed. However,
     /// the behavior of the linear device in that case should be treated as
     /// undefined.
-    /// TODO: If the linear device already exists, verify that the kernel's
-    /// model matches the segments argument.
     pub fn new(name: &DmName, dm: &DM, segments: Vec<Segment>) -> DmResult<LinearDev> {
         if segments.is_empty() {
             return Err(DmError::Dm(ErrorEnum::Invalid,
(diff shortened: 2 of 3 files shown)


mulkieran:
@agrover Please take a look. The testing framework is incomplete and I don't intend to include it in the final PR. But it is handy right now.

agrover:
Seems good so far.

Why do you think device list is more certain, in 34dbe4d?

Reloading the table if the device exists I hadn't thought of... I'll have to think more about this!

mulkieran:
Other than using ```device_list()```, the only options are to try something, and then to assume that an error means the device does not exist. But an error could be caused by lots of reasons. Using ```device_list()``` seems less fuzzy. If there is an error, then we don't know anything. But if there is no error then the name is either in the list or it isn't. All very cut-and-dried.

mulkieran:
About reloading the table...reloading or not reloading are equally fraught with danger, AFAICT. I can not see a way to interrogate DM to discover what the table contains. If there were then, in these ```new()``` or ```setup()``` methods as appropriate, I would instead do a comparison and fail if DM and Stratis view did not match up.

It may be that this is a question to raise with DM.

agrover:
you can get current table by calling dm.table_status with the DM_STATUS_TABLE flag.

mulkieran:
rebased.

mulkieran:
rebased.

mulkieran:
clippy has compilation errors, hopefully they will be fixed soon.

mulkieran:
Making data block size the minimum value gets an I/O error instead of an invalid argument error. Progress.

mulkieran:
@agrover @trgill review?

These are a bunch of disparate changes that are probably mostly correct. They do not completely solve any big problem, but they might as well go up for review now.

src/thindev.rs
```diff
@@ -215,25 +216,24 @@ impl ThinDev {
...
     /// with its thin id from the thinpool.
     pub fn destroy(self, dm: &DM, thin_pool: &ThinPoolDev) -> DmResult<()> {
         let thin_id = self.thin_id;
         self.teardown(dm)?;
-        thin_pool.message(dm, &format!("delete {}", thin_id))?;
-
-        Ok(())
+        thin_pool
+            .message(dm, &format!("delete {}", thin_id))
+            .map(|_| ())
```
trgill: the .map(|_| ()) is necessary here?

mulkieran: I don't think it is. Removing.

mulkieran:
Mad some changes to "Use a final map..." commit.

src/thindev.rs
trgill: FYI - there are a few cases.  I only marked this one.

mulkieran: I looked over everything, the only changes are in "Use a final map..." commit. If there is anything there that you think is wrong, let me know.

agrover:
Please drop the `map(|_| ())`s. They strike me as less clear than what's there now.

Remove `device_exists` check from `new()`. We want `setup` to work if the device already exists but I don't think this applies to `new`.

mulkieran:
"Use expression in place of identifier" substituted for "Use final map...".

mulkieran:
```ThinPoolDev::setup()``` invokes ```ThinPoolDev::new()```, so ```new()``` must work in the case that the device is already known to the kernel in order for ```setup()``` to work.

mulkieran:
@agrover review?

agrover:
> ThinPoolDev::setup() invokes ThinPoolDev::new(), so new() must work in the case that the device is already known to the kernel in order for setup() to work.

Here are the current requirements:

new:
1. doesn't call thin_check
1. errors out if name already exists

setup:
1. calls thin_check
1. doesn't error out if name already exists

Before, they only differed by requirement 1, so implementing one with the other was the simplest thing to do. Now that we have added requirement 2, this may no longer be the case, in which case the implementation must change, not the requirements.

mulkieran:
How have these requirements been arrived at?

agrover:
They were implicit in my comments on this PR. I wanted to explicitly state them since I'm not sure if there's a disagreement over the requirements or the implementation, and we can discuss them separately.

mulkieran:
We would also like ability to teardown meta and data devices if construction of ThinPoolDev fails either in ```setup()``` or ```new()```, possibly. We should consider that complication, while we're worrying about all the rest.

I've pushed off the easily acked commits in this PR to #138.

mulkieran:
rebased.

mulkieran:
rebased.

mulkieran:
rebased.

mulkieran:
rebased.

mulkieran:
```ThinPoolDev::new``` is called exactly when a pool is constructed for the first time. It is not clear whether it should call ```thin_check``` or not. It is reasonable, but maybe optimistic, to expect that invoking ```thin_check``` on a new thinpool would be benign, and might be useful.

trgill:
@mulkieran if we use a recent version of "thin_check" it fails with:

syscall 'open' failed: Device or resource busy
Note: you cannot run this tool with these options on live metadata.
test real_test_pool_setup ... FAILED

It is why the CI tests are failing.

mulkieran:
rebased.

mulkieran:
@agrover In ```ThinPoolDev::new()``` why should an error be returned if the device already exists? Why can't we just continue if we happen to have the correct meta and data devices? Also, this bears on #153 .

mulkieran:
rebased.

agrover:
What's a use case for `new` not returning an error if the thinpool already exists? I can't think of one.

 If the user was setting up an existing thinpool they should've called `setup`. If the user actually is trying to create a new thinpool then returning an existing one will just cause weird failures down the line, because the user needs to track things like used thin_ids, and the assumption when `new` succeeds is that the thinpool is empty/unused, which will likely not be true if an existing thinpool is returned.

mulkieran:
rebased.

mulkieran:
rebased.

mulkieran:
Obsoleted by a bunch of subsequent PRs.

PR #112, opened 2017-07-27
Do not pass dm parameter to table_reload()

mulkieran:
It is already part of the DM implementation, so the correct thing to do
is to use self.


mulkieran:
@trgill review?

mulkieran:
Needs an addition...

mulkieran:
Superceded by #110, since it has become more integral to that PR.

PR #119, opened 2017-07-31
Make new() and setup() methods approach idempotency

mulkieran:
Do not fail, if when setting up or creating a DM device, it is already registered with the kernel.

### DIFF
--- a/src/dm.rs
+++ b/src/dm.rs
@@ -489,21 +489,6 @@ impl DM {
         Ok(DeviceInfo::new(hdr))
     }
 
-    /// Reload the table for a device
-    pub fn table_reload<T1, T2>(&self,
-                                dm: &DM,
-                                id: &DevId,
-                                table: &[TargetLineArg<T1, T2>])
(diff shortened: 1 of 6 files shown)


mulkieran:
This is the non-test code from #110. I plan to put together a test PR separately.
However, the code itself should be identical to that in in #110.

mulkieran:
This code is also being tested indirectly in https://github.com/stratis-storage/stratisd/pull/490, which is currently passing. However, the test in that PR is utterly basic, I welcome suggestions for better tests.

mulkieran:
@agrover @trgill review?

agrover:
Looks fine to me 👍

mulkieran:
@trgill review?

mulkieran:
@agrover @trgill I'll try to come up with better testing on the stratisd side. But if this looks good enough, it might be time to cut a new release. Then we can make use of this change and division through the release instead of the git branch.

PR #122, opened 2017-08-04
Travis: Use container infrastructure

rnestler:
This removes the `sudo: required` from all builds except the `sudo_test` one, which should improve the CI boot time. See https://docs.travis-ci.com/user/reference/precise/#Virtualization-environments for more information.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,11 @@
-sudo: required
 os: linux
 dist: trusty
 group: edge
 
-before_install:
-    - sudo apt-get -qq update
-    - sudo apt-get install -y libdevmapper-dev
+addons:
+  apt:
+    packages:
+    - libdevmapper-dev
 
 language: rust
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,7 @@ matrix:
         - rust: stable
           env: TASK=build
         - rust: stable
+          sudo: required
           env: TASK=sudo_test
         - rust: nightly
           env: TASK=clippy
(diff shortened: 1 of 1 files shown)


rnestler:
> Why is it "Try ...", why not just "Enable..."?

Ah sorry, I was not sure at first if it will work and forgot to rephrase it during rebase.

rnestler:
>Weirdly, though, the fmt target is taking about 3 minutes longer than previously. @rnestler Do you have any idea why?

No idea why, seems odd :confused: I'll rephrase the commit message and re-trigger a build.

rnestler:
We could try to enable [cargo caching](https://docs.travis-ci.com/user/caching/#Rust-Cargo-cache) to further improve build times.

agrover:
👍

mulkieran:
@agrover I'm trying to understand why cargo cacheing would help in our situation (it doesn't seem to hurt).

rnestler:
>@agrover I'm trying to understand why cargo cacheing would help in our situation (it doesn't seem to hurt, but I think we should understand it).

The cargo cache will cache the download and compilation of all dependencies and the rustfmt binary.

I'll re trigger the build to check if there are any issues if the cache is present (cargo may complain if rustfmt is already installed).

rnestler:
Ah the Makefile doesn't try to install rustfmt if it is already present, so it works fine with caching.

mulkieran:
Ok. It all looks fine to me. @agrover if you're satisfied, feel free to merge.

agrover:
Thanks @rnestler!

PR #126, opened 2017-08-07
rustfmt test

mulkieran:


mulkieran:
rustfmt has a complicate new rule about putting commas after the last argument in an argument list. This was intended to investigate whether or not a bug still existed in rust nightly, but it is gone there.

PR #127, opened 2017-08-08
Add a testing infrastructure and some more tests

mulkieran:
This takes the loopbacked testing infrastructure from stratisd and puts it in the src directory. Using this arrangement, it is possible to run tests using external devices using the Rust unit test framework. The PR also add a few more dm tests that don't require the framework, and a bunch of lineardev unit tests that mostly make use of the framework.

I believe that it would be straightforward to also port the "real" testing framework from stratisd, but I have not done so.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,3 +20,7 @@ serde = "1"
 log = "0.3"
 env_logger="0.4"
 clippy = {version = "0", optional = true}
+
+[dev-dependencies]
+loopdev = "0.1.1"
+tempdir = "0.3.5"
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ build:
 	RUSTFLAGS='-D warnings' cargo build
 
 test:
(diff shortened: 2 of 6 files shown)


mulkieran:
I am dissatisfied with the fact that the arguments for many of the methods are ```&[&Path]``` but I've experimented a good deal with types and I don't seem to be able to find any solution that makes use of ```AsRef<Path>```.

mulkieran:
These changes are ported from #110.

mulkieran:
@agrover @trgill review?

mulkieran:
rebased.

trgill:
What do you think about moving LoopTestDev and RealTestDev to a separate crate and both devicemapper-rs and stratisd use that crate?

mulkieran:
@trgill Maybe we should. I had some reservations about the idea that made me prefer this approach, but they weren't very compelling. We should talk about it in meeting.

PR #131, opened 2017-08-09
Do not cache

mulkieran:


mulkieran:
This was a test. It does not seem that cacheing actually has any effect on the version of clippy that is run, which is good.

PR #136, opened 2017-08-15
implement snapshot for a thindev

trgill:
Taking a snapshot of a ThinDev creates a new ThinDev that is no different
than any other thin provisioned device.

Documentation for snapshots can be found at:

https://www.kernel.org/doc/Documentation/device-mapper/thin-provisioning.txt

See section i) Creating an internal snapshot.

### DIFF
--- a/src/thindev.rs
+++ b/src/thindev.rs
@@ -370,6 +370,32 @@ mod tests {
         tp.teardown(&dm).unwrap();
     }
 
+    /// Verify success when taking a snapshot of a ThinDev.  Check that
+    /// the size of the snapshot is the same as the source.
+    fn test_snapshot(paths: &[&Path]) -> () {
+        assert!(paths.len() >= 1);
+        let td_size = MIN_THIN_DEV_SIZE;
+        let dm = DM::new().unwrap();
+        let tp = minimal_thinpool(&dm, paths[0]);
+
+        // Create new ThinDev as source for snapshot
+        let thin_id = ThinDevId::new_u64(0).expect("is below limit");
+        let thin_name = DmName::new("name").expect("is valid DM name");
+        let td = ThinDev::new(&thin_name, &dm, &tp, thin_id, td_size).unwrap();
+
+        // Create a snapshot of the source
+        let ss_id = ThinDevId::new_u64(1).expect("is below limit");
+        let ss_name = DmName::new("snap_name").expect("is valid DM name");
+        let ss = td.snapshot(&dm, &tp, ss_name, ss_id).unwrap();
+
+        // Verify the source and the snapshot are the same size.
+        assert_eq!(td.size(), ss.size());
+
+        ss.destroy(&dm, &tp).unwrap();
+        td.destroy(&dm, &tp).unwrap();
+        tp.teardown(&dm).unwrap();
+    }
+
     #[test]
     fn loop_test_basic() {
         test_with_spec(1, test_basic);
@@ -383,5 +409,9 @@ mod tests {
     #[test]
     fn loop_test_setup_without_new() {
         test_with_spec(1, test_setup_without_new);
+    }
+    #[test]
+    fn loop_test_snapshot() {
+        test_with_spec(1, test_snapshot);
     }
 }
--- a/src/thindev.rs
+++ b/src/thindev.rs
@@ -9,7 +9,7 @@ use serde;
 
 use super::device::Device;
 use super::deviceinfo::DeviceInfo;
-use super::dm::{DM, DevId, DmFlags, DmName};
+use super::dm::{DM, DM_SUSPEND, DevId, DmFlags, DmName};
 use super::result::{DmError, DmResult, ErrorEnum};
 use super::shared::{DmDevice, device_create, device_exists, device_setup, table_reload};
 use super::thinpooldev::ThinPoolDev;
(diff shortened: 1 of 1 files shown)


trgill:
@mulkieran @agrover review please?

agrover:
Why 5a977ce? Device is `Copy` so it shouldn't be necessary, is my thought.

trgill:
@agrover you are right.  I removed the commit.

src/thindev.rs
```diff
@@ -184,6 +184,36 @@ impl ThinDev {
            })
     }
 
+    /// Create a snapshot of a ThinDev.  Once created a snapshot
+    /// is the same as any other thin provisioned device.  There is
```
mulkieran: Does it make more sense to define this method on a ThinPoolDev? Like:

```
pub fn snapshot(&self, dm: &DM, origin: &DevId, origin_thin_id: ThinDevId, snapshot_name: &DmName, snapshot_thin_id: ThinDevId) {
  dm.device_suspend(origin, ...);
  self.message(dm, &format(..., snapshot_thin_id, origin_thin_id));
  dm.device_suspend(origin, ...);
  let dev_info = device_create(dm, snapshot_name, &ThinDev::dm_table(self.device(), snapshot_thin_id, <maybe size would have to be an addtional param>);
  Ok(...)
}
```

trgill: I was thinking snapshot() is similar to ThinDev::new() and ThinDev::setup() since it returns a new ThinDev.

Why are you thinking it belongs in ThinPoolDev?  Because we message the ThinPoolDev?

agrover: honestly I could see it either way...

mulkieran: There is at least one significant difference. In ThinPoolDev snapshot method it is necessary to pass the size of the snapshot. But I think that for a (theoretically general purpose) library, that might be a good thing. What happens if snapshot size > origin size, or vice-versa?

Also, parameters are a bit less odd, in my version, I think.

trgill: Snapshot size is always equal to the source size - they are identical at the time of the snapshot.

I'll implement what @mulkieran suggested and see how it works.  Will report back.

mulkieran: There are two things that happen in the method:

1. The thin pool gets informed of a snapshot, and gets the relation via two thin ids, one the snapshot's and one the origins.
2. The snapshot device is created and the size is specified. This is just another thin device with the snapshot id and a specified size. What would happen if you specified a larger size at this time? Would table_load fail? Would it succeed, and if it did, what would be the consequences?

trgill:
So, this doesn't compile if I move it to thinpooldev.rs - the fields are private.

```
        Ok(ThinDev {
               dev_info: dev_info,
               thin_id: snapshot_thin_id,
               size: self.size(),
               thinpool: self.device(),
           })
```

I'll add a new method to thindev.rs to construct a ThinDev given these fields?

trgill:
@mulkieran yeah, I've got a test in a stratisd PR.

mulkieran:
@trgill I wasn't clear. In devicemapper-rs, there is now an existing loopbacked based testing infrastructure. In both thinpool.rs and thinpooldev.rs there are testing modules w/ some tests already written. You should be able to throw in an additional test for this new function, regardless of where it ends up, either as part of ThinDev or as part of ThinPoolDev.

mulkieran:
@trgill Maybe you are better off putting it in ThinPool. But you could still change the parameters as:

```
pub fn snapshot(&self,
                           dm: &DM,
                           thin_pool: &ThinPoolDev,
                           snapshot_name: &DmName,
                           snapshot_thin_id: ThinDevId,
                           size: Sectors) { // size might be optional, defaulting to self.size() ?
  let source_id = DevId::Name(self.name());
  dm.device_suspend(&source_id, ...);
  thin_pool.message(dm, &format(...));
  dm.device_suspend(&source_id, ...);
  let table = ThinDev::dm_table(thin_pool.device(), snapshot_thin_id, size);
  let dev_info = device_create(dm, snapshot_name, &table);
  Ok(...)
}
```
Seems to bring it into line a bit more w/ other ThinDev methods.

trgill:
@mulkieran I updated the parameters to ThinDev::snapshot() as you suggested.

I also added a very simple unit test that just compares the snapshot size with the source.  I could enhance to write some files - but my plan was to do that as an integration test.

src/thindev.rs
```diff
@@ -184,6 +184,35 @@ impl ThinDev {
...
+    /// snapshot.
+    pub fn snapshot(&self,
+                    dm: &DM,
+                    thin_pool: &ThinPoolDev,
+                    snapshot_name: &DmName,
+                    snapshot_thin_id: ThinDevId)
+                    -> DmResult<ThinDev> {
+        let source_name = DevId::Name(self.name());
+        dm.device_suspend(&source_name, DM_SUSPEND)?;
+        dm.target_msg(&DevId::Name(thin_pool.name()),
```
mulkieran: You can use ```thin_pool.messag(...``` here.

src/thindev.rs
```diff
@@ -184,6 +184,35 @@ impl ThinDev {
...
+    /// is the same as any other thin provisioned device.  There is
+    /// no need to track any connection between the source and the
+    /// snapshot.
+    pub fn snapshot(&self,
+                    dm: &DM,
+                    thin_pool: &ThinPoolDev,
+                    snapshot_name: &DmName,
+                    snapshot_thin_id: ThinDevId)
+                    -> DmResult<ThinDev> {
+        let source_name = DevId::Name(self.name());
```
mulkieran: Minor, but this really is an id, not a name.

trgill: Ok

src/thindev.rs
trgill: Ok

trgill:
@mulkieran 
@agrover maybe you know the answer?

I don't think I said a snapshot couldn't be made larger or smaller than the origin.  The documentation shows the snapshot table being the same as the origin.

The following is from linux/drivers/md/dm-thin.c at ~line 53

 ```
* When you take an internal snapshot you clone the root node
 * of the origin btree.  After this there is no concept of an origin or a
 * snapshot.  They are just two device trees that happen to point to the
 * same data blocks.

```

How about I open an issue to investigate?  I don't think we want to hold this PR.

trgill:
I opened issue #177 to track concerns about changing size of snapshot table.

trgill:
rebased

agrover:
👍

PR #139, opened 2017-08-18
Some changes to the definition and use of DevId

mulkieran:


mulkieran:
This is in progress because ```DM_UUID_LEN``` is a puzzle. Its value is 129. But a raw UUID is an 128-bit number. This means that the expectation is that the UUID will take up one byte for every bit that it actually requires. This _seems_ fishy.

Note that ```DM_NAME_LEN``` is 128. This means that names can be 127 characters plus a null terminator, which is generous.

agrover:
Here's some output on my system. It looks like both name and uuid are big, unique text strings, so UUID field isn't actually a 128 bit value. Huh.

```
luks-d874ad94-8d7e-4134-8ba3-a7677b2651ef: 0 775946240 crypt aes-xts-plain64 00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000 0 8:2 4096
luks-13bf9bd7-a488-4143-9462-c04b6b987bcc: 0 4194304 crypt aes-xts-plain64 00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000 0 8:3 4096
[agrover@work ~]$ sudo dmsetup info luks-d874ad94-8d7e-4134-8ba3-a7677b2651ef
Name:              luks-d874ad94-8d7e-4134-8ba3-a7677b2651ef
State:             ACTIVE
Read Ahead:        256
Tables present:    LIVE
Open count:        1
Event number:      0
Major, minor:      253, 0
Number of targets: 1
UUID: CRYPT-LUKS1-d874ad948d7e41348ba3a7677b2651ef-luks-d874ad94-8d7e-4134-8ba3-a7677b2651ef
```

mulkieran:
Yes. I finally remembered all that. It doesn't make that much difference to the code, but we lose all sorts of type-safety benefits.

PR #140, opened 2017-08-18
Use DevId a bit more and document it a bit more

### LINKED ISSUE #90: We know that the ioctl call for device_rename does not update the appropriate field to the new name
We don't know whether this is a bug or a feature or how to find out. We don't know whether the same occurs for UUIDs, but that is easy to find out. We might look at dmlib for some hints.

### LINKED ISSUE #92: Make device_rename DeviceInfo have correct info for name
The rename DM ioctl doesn't update the name field in the Struct_dm_ioctl it returns.

This is important since calling `Lineardev::set_name` will result in the LinearDev thinking it still has the old name. (It just saves the returned Struct_dm_ioctl.) Future operations will thus fail. Possible solutions:

1. Change DM to fill in the new name
1. Change dm::device_rename to fill in the new name in the to-be-returned Struct_dm_ioctl if the ioctl succeeded
1. Change dm::device_rename to call device_status afterwards (with the new name), and then returning the fresh Deviceinfo thus obtained
1. Modify DeviceInfo to not simply contain a Struct_dm_ioctl
1. Have LinearDev::set_name call device_status for a correct DeviceInfo
1. Have LinearDev::set_name call device_rename again to get a correct DeviceInfo

mulkieran:
* Documents the reason why DevId's Uuid constructor does not take a Uuid argument.
* Rewrite ```device_rename()``` to use DevId parameter. Fixes a bug since it is impossible to reset a uuid according to kernel docs in ioctl.h.
* Changes a confusing parameter name, "name", to "id", which is a bit better.
* Adds some tests to verify basic operation on devicemapper "uuids".
* Has ```LinearDev::set_name()``` return the DeviceInfo from ```device_status()``` call.
* Makes ```LinearDev::set_name()``` idempotent.

Resolves #90.
Resolves #92.

### DIFF
--- a/src/dm.rs
+++ b/src/dm.rs
@@ -32,7 +32,9 @@ use util::slice_to_null;
 pub enum DevId<'a> {
     /// The parameter is the device's name
     Name(&'a str),
-    /// The parameter is the device's UUID
+    /// The parameter is the device's "DM UUID"
+    /// Note that this UUID is not a canonical UUID, but rather a
+    /// devicemapper-specific format that is supposed to be unique.
     Uuid(&'a str),
 }
 
@@ -69,6 +71,8 @@ impl DM {
         name_dest[..len].clone_from_slice(name.as_bytes());
     }
 
+    /// Set the devicemapper "UUID".
+    /// Note that this is a devicemapper specific id value, not a UUID.
     fn hdr_set_uuid(hdr: &mut dmi::Struct_dm_ioctl, uuid: &str) -> () {
         let uuid_dest: &mut [u8; DM_UUID_LEN] = unsafe { transmute(&mut hdr.uuid) };
         let len = uuid.as_bytes().len();
@@ -225,7 +229,7 @@ impl DM {
     /// use devicemapper::consts::DmFlags;
     /// let dm = DM::new().unwrap();
     ///
-    /// // Setting a uuid is optional
+    /// // Setting a devicemapper uuid is optional
     /// let dev = dm.device_create("example-dev", None, DmFlags::empty()).unwrap();
     /// ```
     pub fn device_create(&self,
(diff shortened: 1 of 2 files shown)


mulkieran:
Supercedes #139 .

mulkieran:
Still wrong...you can never reset the uuid.

mulkieran:
@agrover review?

mulkieran:
Actually, why don't I implement choice (5) from #92 first.

mulkieran:
@agrover review?

mulkieran:
Not as wrong as before, but still wrong. It should be possible to identify the device with a uuid and set its name to some new value.

One thing we can and should do, but it won't help much, is to establish type aliases, as:

```
type DmName = str;
type DmUuuid = str;
```
They won't help with type safety, but they will indicate intention a bit better.

We shouldn't try to achieve idempotency in DM methods, because there is no effort to in the underlying ioctls and because the choice to return the possibly altered DeviceInfo from the ioctl call in the methods, means that achieving idempotency in the methods would require faking up DeviceInfo values, which would lead to all sorts of confusion and headache.

Dealing w/ ```device_rename()``` is still an open question.

* Split it into two functions, one to set the uuid and one to reset the name.
```
device_rename(&self, id: &DevId, new_name: &DmName) -> DmResult<DeviceInfo> ...
device_set_uuid(&self, name: &DmName, uuid: &DmUuid) -> DmResult<DeviceInfo> ...
```
At least this allows only meaningful operations. Which function is called is what decides whether or not the ```DM_UUID``` flag is passed to the ioctl call.

* Keep ```device_rename()```
```
device_rename(&self, old: &DevId, new: &DevId) -> DmResult<DeviceInfo> ...
```
This allows the meaningless operation of using the uuid to select the device and also specifying a new uuid, which should always fail. Whether or not the ```DM_UUID``` flag is passed to the ioctl call is decided by the constructor of ```new```.

* Other ideas...

mulkieran:
Attempting the first in #141.

mulkieran:
Superceded by #141.

PR #141, opened 2017-08-21
Work on devicemapper identifiers

### LINKED ISSUE #90: We know that the ioctl call for device_rename does not update the appropriate field to the new name

### LINKED ISSUE #92: Make device_rename DeviceInfo have correct info for name

mulkieran:
This PR is supposed to help clarify the use of various things for identifying devices.

* It rewrites DevId substantially.
* It documents that a devicemapper uuid is not an ISO standard UUID.
* It changes a confusing parameter name, "name", to "id", which is a bit better.
* ```DM::device_rename()``` does not allow specifying the device by uuid either when renaming or when setting a uuid.
* Has LinearDev::set_name() return the DeviceInfo from device_status() call.
* Makes LinearDev::set_name() idempotent.
* Adds a bunch of tests.

Resolves #90.
Resolves #92.

### DIFF
--- a/src/deviceinfo.rs
+++ b/src/deviceinfo.rs
@@ -45,21 +45,19 @@ impl DeviceInfo {
     }
 
     /// The device's name.
-    pub fn name(&self) -> &DmName {
+    pub fn name(&self) -> DmName {
         let name: &[u8; DM_NAME_LEN] = unsafe { transmute(&self.hdr.name) };
-        // no chance not null-terminated
-        let slc = slice_to_null(name).unwrap();
-        // no chance this isn't utf8 (ascii, really)
-        from_utf8(slc).unwrap()
+        let slc = slice_to_null(name).expect("kernel ensures null-terminated");
+        let name = from_utf8(slc).expect("kernel ensures ASCII characters");
+        DmName::new(name).expect(".len() < DM_NAME_LEN")
     }
 
-    /// The device's UUID.
-    pub fn uuid(&self) -> &DmUuid {
+    /// The device's devicemapper uuid.
+    pub fn uuid(&self) -> DmUuid {
         let uuid: &[u8; DM_UUID_LEN] = unsafe { transmute(&self.hdr.uuid) };
-        // no chance not null-terminated
-        let slc = slice_to_null(uuid).unwrap();
-        // no chance this isn't utf8 (ascii, really)
-        from_utf8(slc).unwrap()
+        let slc = slice_to_null(uuid).expect("kernel ensures null-terminated");
+        let uuid = from_utf8(slc).expect("kernel ensures ASCII characters");
+        DmUuid::new(uuid).expect(".len() < DM_UUID_LEN")
     }
 
     /// The flags returned from the device.
--- a/src/deviceinfo.rs
+++ b/src/deviceinfo.rs
@@ -8,6 +8,7 @@ use dm_ioctl as dmi;
 
 use consts::{DM_NAME_LEN, DmFlags, DM_UUID_LEN};
 use device::Device;
+use dm::{DmName, DmUuid};
 use util::slice_to_null;
 
 /// Contains information about the device.
@@ -44,7 +45,7 @@ impl DeviceInfo {
     }
 
     /// The device's name.
-    pub fn name(&self) -> &str {
+    pub fn name(&self) -> &DmName {
         let name: &[u8; DM_NAME_LEN] = unsafe { transmute(&self.hdr.name) };
         // no chance not null-terminated
         let slc = slice_to_null(name).unwrap();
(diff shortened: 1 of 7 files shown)


mulkieran:
Supercedes #140.

mulkieran:
@agrover review?

This is the best I can do. I can't find any explanation for why you can't set a new name on a device that you have specified by a uuid, but it fails consistently. Other things where you specify a uuid succeed consistently, so there is nothing simple going on.

The kernel source code and the kernel docs and the devicemapper-rs tests agree that you can not reset a uuid. On pretty solid ground, there.

agrover:
I think this is a limitation (bug?) in the kernel. See dm-ioctl.c. Other DM functions use `__find_device_hash_cell` (possibly via `find_device`), which checks the first present of (uuid, name, devno). However, when searching for "old", `dm_hash_rename` doesn't call that, it only looks up the name with `__get_name_cell`.

That's the reason. It's hard to know if not calling `__find_device_hash_cell` in `dm_hash_rename` is an oversight or if there is some subtle reason it couldn't, but at this point it's probably moot.

I'll look at the rest of this PR tomorrow.

mulkieran:
Updated ```device_rename()``` comment and commit message.

mulkieran:
rebased.

agrover:
I'm not convinced `DmNameRef` and friends are quite right. I think we might be happier if we went further and actually defined new types (not just aliases). This would help with clarity like the `type` approach, but also let us only construct valid DM names/uuids by checking max length & allowed characters, which then could be passed to other methods that wouldn't need to recheck validity.

Also suggest renaming to DmName, DmNameBuf, DmUuid, and DmUuidBuf to line up with other types like Path/PathBuf (although the stdlib certainly isn't consistent about names for ref/owned types  - OsStr/OsString, CStr/CString, and of course str/String...)

Rest of this PR is 👍 ! I'm fine with either dropping d928aa2 and merging the rest now and do a followup PR for the additional work described above, or holding and revising this PR.

mulkieran:
Renamed type aliases.

mulkieran:
I think the types for DmUuid and DmName should be as minimal as possible. We don't really want to duplicate what devicemapper itself does, especially if we don't really know what it is. But, we need to know the maximum lengths of both in order to do operations in devicemapper-rs, so I think it makes sense to incorporate a type that only allows values of the maximum allowable length.

mulkieran:
@agrover review please?

The new matter is all in the last 4 commits, which implement ```DmUuid``` and ```DmName``` as types. My plan is to merge them into a single commit.

My other plan is to macroize ```DmName``` and ```DmUuid``` since they are virtually identical. But I'ld like to get your feedback on what I've done so far, before I head off into that.

agrover:
Hi, looks good. I'd recommend implementing [Deref as well](https://doc.rust-lang.org/book/second-edition/ch15-02-deref.html#implicit-deref-coercions-with-functions-and-methods) for each pair. And then, maybe try converting stratisd to use the new API as a test to see how easy it is to use and how involved the changes over there will be?

mulkieran:
@agrover, review? I plan to squash the last four into one commit.

mulkieran:
@agrover I'm rethinking getting rid of DmNameBuf...

agrover:
so far looks good to me, yes.

mulkieran:
For reference: https://github.com/stratis-storage/stratisd/pull/540.

mulkieran:
@agrover review, please? I plan to squash the last five commits. There's a link above to a PR containing the necessary changes in stratis.

agrover:
CI needs a look but 👍 from me.

mulkieran:
Ok, squashed and fixed some bad formatting tests.

lilianmoraru:
This broke the API(running `cargo update` on `stratisd` and then trying to build), probably needed a major version bump(`0.6.2` -> `0.7.0`).

PR #144, opened 2017-08-25
Change DmName/DmUuid to be unsized

agrover:
A little poking around to see if we could implement ToOwned for DmName has led me to the conclusion that DmName should be an unsized type, wrapping `str` not `&str`.

This PR undoes some changes in #141. In looking at the code more I gained some insights that would have been REALLY good to bring up in the earlier PR's review. I'm sorry for not doing so earlier and noticing before it was merged (and tagged).

### DIFF
--- a/src/dm.rs
+++ b/src/dm.rs
@@ -3,6 +3,7 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 use std::ascii::AsciiExt;
+use std::borrow::Borrow;
 use std::fmt;
 use std::fs::File;
 use std::io::Error;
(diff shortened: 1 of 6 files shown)


mulkieran:
The failing tests should be fixed...

src/dm.rs
```diff
@@ -47,92 +48,96 @@ fn dev_id_check(value: &str, max_allowed_chars: usize) -> DmResult<()> {
...
+            fn borrow(&self) -> &$B {
+                &self.as_ref()
+            }
+        }
 
-/// The owned version of DmName.
-#[derive(Debug, PartialEq, Eq)]
-pub struct DmNameBuf {
-    value: String,
+        impl Deref for $O {
```
mulkieran: It looks like this implementation is the context where a new $B should be created. It seems like ```AsRef```, ```Borrow```, and ```Deref``` are kind of exchanging their usual roles, here.

src/dm.rs
```diff
@@ -382,7 +387,7 @@ impl DM {
     /// must be "".
     /// Note: Possibly surprisingly, returned DeviceInfo's uuid or name field
     /// contains the previous value, not the newly set value.
-    pub fn device_rename(&self, old_name: DmName, new: DevId) -> DmResult<DeviceInfo> {
+    pub fn device_rename(&self, old_name: &DmName, new: DevId) -> DmResult<DeviceInfo> {
```
mulkieran: Think you missed ```device_remove```, ```device_status```, ```table_clear```, maybe a few others. Worth taking a look at orignal commit.

mulkieran:
Since this is a complete rewrite of DmName, DmUuid, DevId, might as well squash everything into one commit, when ready.

mulkieran:
You should also take over in what ever way is most convenient, this PR: https://github.com/stratis-storage/stratisd/pull/540. It makes the necessary changes in stratisd to keep up with the changes in the API.

It would be good to have that fully prepared before merging and tagging this PR, so it can be merged immediately.
It looks like we should make that a policy in future, i.e., that we always have a stratisd update for new devicemapper version in the queue before we tag the devicemapper version.

mulkieran:
Maybe you should still be passing ```DevId``` parameters by value. That continues to make sense to me, because DevId is still just a tag and a reference to a Dm*.

mulkieran:
Blocks #137.

agrover:
squashed into one commit, made changes based on review feedback and to pass CI.

I'm leaving `DevId` params refs for now, since I'm still not sure.

src/dm.rs
```diff
@@ -47,92 +48,102 @@ fn dev_id_check(value: &str, max_allowed_chars: usize) -> DmResult<()> {
...
-        impl<'a> AsRef<str> for $T<'a> {
+        impl AsRef<str> for $B {
             fn as_ref(&self) -> &str {
-                self.value
+                &self.inner
             }
         }
 
-        impl<'a> fmt::Display for $T<'a> {
+        impl Deref for $B {
```
mulkieran: Can you omit, in interests of conssitency with Path?

src/dm.rs
```diff
@@ -47,92 +48,102 @@ fn dev_id_check(value: &str, max_allowed_chars: usize) -> DmResult<()> {
...
+        }
 
-/// The owned version of DmName.
-#[derive(Debug, PartialEq, Eq)]
-pub struct DmNameBuf {
-    value: String,
+        impl Deref for $O {
+            type Target = $B;
+            fn deref(&self) -> &$B {
+                $B::new(&self.inner).unwrap()
```
mulkieran: Omit needless unwraps()! ```$B { inner: &self.inner }``` should work.

src/dm.rs
agrover: We need to get the `&str` in the impl, we call `as_bytes()` on it. We can add an `as_str()` method on $B?

src/dm.rs
agrover: That syntax doesn't work. Had to do `unsafe { transmute(self.inner.deref()) }`.

agrover:
Ok, two new to-be-squashed commits for most recent review comments.

mulkieran:
Ok. Let's go back to implementing Deref for $0 w/ new, but use an expect() instead.

```
        impl Deref for $O {
            type Target = $B;
            fn deref(&self) -> &$B {
                $B::new(&self.inner).expect("inner satisfies all correctness criteria for $B::new")
            }
        }
```

also, instead of implementing ```as_str()```, and then calling ```as_bytes()``` on the result, why not just call ```as_bytes()``` on the result of ```as_ref()```, which you have anyway?

Also, in your header comments on ```dev_id``` macro, it would probably be helpful to indicate that the implementation is intended to resemble as closely as is sensible the implementation of Path/PathBuf.

mulkieran:
Btw, I tried to implement ```AsRef<DevId>``` for ```DmName```. I thought that would be handy, so that a ```DmName``` could be passed where a ```DevId``` was wanted. But it seems the lifetimes can't be satisfied. Then I thought one could just implement an ```as_devid``` method for ```DmName```, but there is little point in that, it just doesn't save any trouble, overall.

agrover:
fixed and squashed, except I didn't quite follow what you were saying about `as_bytes` so no changes there.

agrover:
rebased

PR #146, opened 2017-08-29
Tidy some mentions of Stratis in devicemapper-rs

agrover:
Devicemapper stands on its own. There are a few harmless mentions of
Stratis in the code and test strings, but let's try to avoid these just on
principle.

Take the opportunity to describe in more detail how we can be sure the DM
kernel interface won't change under us and cause a crash.

### DIFF
--- a/src/dm.rs
+++ b/src/dm.rs
@@ -1,6 +1,7 @@
 // This Source Code Form is subject to the terms of the Mozilla Public
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
 use std::ascii::AsciiExt;
 use std::fmt;
 use std::fs::File;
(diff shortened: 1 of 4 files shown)


src/dm.rs
```diff
@@ -641,10 +642,12 @@ impl DM {
...
     /// canonicalization makes checking identity of tables easier.
-    // The justification for this is that if there was no error in obtaining
-    // the table the data is correct and complete. Therefore, an error in
-    // parsing can only result from a change in the kernel. We assume that
-    // Stratis will know in advance about such changes.
+    // Justification: If the ioctl succeeded, the data is correct and
+    // complete. An error in parsing can only result from a change in the
+    // kernel. We rely on DM's interface versioning system. Kernel changes
+    // will either be backwards-compatible, or will increment
+    // DM_VERSION_MAJOR.  Since calls made with an non-matching major version
```
mulkieran: "an non..." -> "a non..."

agrover:
typo fixed

PR #150, opened 2017-09-01
Clean up DM device if table load fails

agrover:
In ThinDev and ThinPoolDev::setup(), we create and load a device if it
doesn't exist. However, if table load fails (e.g. if attempting to load
a thindev whose thin_id doesn't exist) then the newly created device is
left dangling. In this case, make sure to not leave the created-but-not-
loaded table as we return with the Err.


agrover:
good idea! Changes made.

agrover:
fixed clippy errors.

PR #151, opened 2017-09-05
Run cargo-tree from Makefile using "make tree"

mulkieran:
This assumes that cargo-tree has been installed in the same way as
cargo-fmt.


mulkieran:

I have often wanted to know all the dependencies of our projects, and this gives a nice few of that information.

I ran into it because I was looking at Rust implementations of DAGs.

PR #152, opened 2017-09-05
Require device() instead of dstr() in DmDevice

mulkieran:
Device is copy, and a better choice as an identifier when not passing the
entire DmDevice as an argument.

ThinDev now stores its thinpooldev as a Device, not a String.
The need for a special type for Dstr goes away entirely, since Device
implements Display in the necessary format.


mulkieran:
@agrover @trgill review?

mulkieran:
rebased.

PR #154, opened 2017-09-06
Conditionally run thin_check on thin pool meta data device

mulkieran:
"thin_check" fails if the meta data device is "busy"; we believe that it is
certainly "busy" if the kernel knows about it.


mulkieran:
I believe that this change, cherry-picked from #110 , is non-controversial, so let's merge it.

@agrover @trgill review?

PR #160, opened 2017-09-08
Remove redundant parameter from device_create

mulkieran:
This was actually a bit of a bug, because it would be bad news if the
name and the id specified different devices.


mulkieran:
@agrover review?

PR #163, opened 2017-09-12
Use error-chain library

mulkieran:


mulkieran:
See Stratis usage at: https://github.com/stratis-storage/stratisd/pull/564.

agrover:
How about renaming `devicemapper::Result` back to `DmResult` using `#[error_chain(result = "DmResult")]`? When I see `Result` I think `std::result::Result`.

mulkieran:
I think that would be an unusual thing to do. even the standard library has ```fmt::Result```, ```thread::Result```, ```io::Result```, ```alloc::fmt::Result```, ```collections::fmt::Result``` and in my sample of (1), that's how dependent crates behave.

agrover:
This isn't how we're doing it right now, either here or in stratisd. It would be best to introduce error chain in this PR without tying it to the larger style issue of how we name our result types, please.

mulkieran:
@agrover @trgill Here's an example backtrace that we get from an unwrap() when the error is an error-chain error and BACKTRACE=1. https://travis-ci.org/stratis-storage/devicemapper-rs/jobs/277303412. Note that it contains the full backtrace of the actual error, although there is also a lot of other probably unavoidable stuff mixed in with it.

mulkieran:
clippy is temporarily failing...

mulkieran:
It would be rational to make up a separate error heirarchy for dm.rs, as it is really the core part. That would move IoctlError, ContextInitError into dm errors, and copy InvalidArgument error. Then, it would be possible to expand the dm error heirarchy, while still matching the Dm error at the top level.

Unfortunately, that would mean exporting two types of errors at the top-level, since stratisd, e.g., must get devicemapper context. Whoops.

mulkieran:
Some notes:

Things we want from error-chain and whether we can get them:
1. A stack trace that tells us more about an error. When we unwrap() a value we get the stack trace that got us to the unwrap(). We would also like to see the stack trace at the point the error was returned. error-chain provides a stack trace of this kind for _every_ error in its chain. (All this depends on how BACKTRACE is set.) error-chain is very satisfactory on this point, and we can get this regardless of whether we actually chain errors or not. On the other hand, if we chain a lot of errors, we can get more backtraces than we really wanted to look at. The documentation does not make it clear whether they would all be different or all the same. It makes sense that they should all be different, as it doesn't make sense to store multiple copies of things, all guaranteed to be the same.
2. We want to be able to extend our errors while retaining backwards compatibility. We might want to add an error, a new constructor to an already existing list of constructors, or we might want to split an existing single error into two errors. In a class-based object-oriented language w/ inheritance, the first would involve adding a new subtype of some existing type, the second adding two subtypes that never previously existed. In either case, code not interested in the new type of error could continue to ignore it, while code requiring the change could do an update to the new version w/ the change. An addition of something new to an existing list is easy to do by adding a new constructor to ```errors {}```. error_chain already forces non-exhaustive matching in clients with the hidden variants hack and may update to new annotation based approach: https://github.com/rust-lang-nursery/error-chain/issues/222.
The problem of splitting an error in two is a good deal harder. One possibility is to use chaining and imagine that it is possible to always distinguish between errors by their cause. Another might be to use links to separately defined error types for any error that _might_ later be divided. Each separate error type need contain no defined constructors, only the default Msg constructor. Since it is the only constructor, it is pointless to match on it, although this can not be forbidden. But in any case, through non-exhaustive matching, the opportunity to add additional constructors is always available. I think that this is reasonable.

mulkieran:
Taking advantage of #171 .

mulkieran:
rebased.

mulkieran:
rebased.

agrover:
@mulkieran this PR appears to be obsoleted, if so please close?

mulkieran:
It needs to be broken out into issues.

mulkieran:
Likely we will be using faillure crate instead of error-chain, so probably this PR is no longer useful.

Issue #72, opened 2017-06-19
Need to clarify rational by which names are exported in public interface


mulkieran:
Should other things be in submodules besides consts?

PR #165, opened 2017-09-12
Disperse constants that are not broadly shared into appropriate modules

mulkieran:
Resolves: #72.


mulkieran:
Note: currently IEC is defined in both devicemapper-rs and stratisd. We should eliminate it from one or the other, probably from stratisd.

mulkieran:
rebased.

mulkieran:
@agrover @trgill review?

mulkieran:
clippy has caught up w/ new release. So, this should succeed now.

mulkieran:
@agrover review?

PR #170, opened 2017-09-20
Pass devicemapper context to LinearDev::extend()

mulkieran:
There is no other method in LinearDev, ThinDev, ThinPoolDev where the
context is instantiated within the method. If the context is required,
it is always passed as a parameter. This makes it the same for this method.


mulkieran:
@agrover @trgill review?

agrover:
seems 👍

PR #172, opened 2017-09-20
Update thindev's size only after table_reload succeeds

mulkieran:
We presume that if the table reload fails, the kernel still considers
size of thindev to be old size, so stratisd should, too.


mulkieran:
@agrover review?

agrover:
👍

PR #176, opened 2017-09-22
What happens if you invert the order of creation and message

mulkieran:


mulkieran:
Inscrutable bad things. I guess we got the order right.

Issue #16, opened 2017-04-12
It is probably a bad idea to have logging code affect non-logging code


mulkieran:
By this, I mean that it would be bad to have changes in the types or other characteristics of names be required by logging code. This might then result in changes to the API, etc. This is more likely in a language like Rust then it would ever be in Python or C.

PR #178, opened 2017-09-27
Omit debug! in LinearDev::dm_table

mulkieran:
I believe that it has served its purpose as a development aid, and it
would be unlikely to be all that helpful for customer debugging.

When there is a failure, we would be better served by packaging that
information up in the error returned.

Get rid of log, env_logger requirements. They can be re-added w/ a small
amount of effort as necessary.

Resolves: #16.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,8 +17,6 @@ bitflags = "1"
 newtype_derive = "0.1"
 macro-attr = "0.2.0"
 serde = "1"
-log = "0.3"
-env_logger="0.4"
 clippy = {version = "0", optional = true}
 error-chain = "0.11"
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -58,9 +58,6 @@ extern crate bitflags;
 #[macro_use]
 extern crate error_chain;
 
-#[macro_use]
-extern crate log;
-
 #[cfg(test)]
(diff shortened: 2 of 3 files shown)


mulkieran:
@agrover @trgill review?

trgill:
I'm not sure I understand the minimalist approach to logging.  If you both agree this is good, please merge. 👍 

We may want to write some scripts to collect all of what we might need to debug from a system.  I suppose DM setup will give us all the table information, but it won't tell us the order of things like a log would.

mulkieran:
rebased.

mulkieran:
rebased.

mulkieran:
rebased.

mulkieran:
rebased.

mulkieran:
rebase + clippy test

mulkieran:
rebased.

mulkieran:
Jenkins test passed but there was a plugin failure.

mulkieran:
rebased.

mulkieran:
Jenkins failure is due to a cleanup problem.

mulkieran:
rebased.

agrover:
OK, let's clear this off the decks and merge it. When we have a more cohesive plan for logging we can trivially walk this back if we decide that is the right thing to do.

PR #182, opened 2017-09-28
Dm events

agrover:


mulkieran:
The only dm tests that are passing are ones that expect an error; they're probably getting a different error than they usually get, at some level.

src/dm.rs
```diff
@@ -378,8 +379,47 @@ impl DM {
                 let slc = slice_to_null(&result[size_of::<dmi::Struct_dm_name_list>()..])
                     .expect("Bad data from ioctl");
                 let dm_name = String::from_utf8_lossy(slc).into_owned();
+
+                // In minor version 36 and later, we now get each device's
```
mulkieran: Please leave out "now", it'll get stale.

src/dm.rs
```diff
@@ -378,8 +379,47 @@ impl DM {
...
+                // event number after its name.
+                // Should match offset calc in kernel's
+                // drivers/md/dm-ioctl.c:list_devices
+                let event_nr = {
+                    match hdr.version[1] {
```
mulkieran: Different structure to isolate diff between 36 and subsequent:

```
match hdr.version[1] {
  0...35 => None,
  ver => {
    // A bug in minor_version 36 was corrected in subsequent versions.
    // Link if there is any.
    // This bug forces a different calculation of offset value.
    let offset = if ver == 36 {
       ...
    } else {
        12 + slc.len() + 1
    }
    
    let aligned_offset ...
    ...
  }
}
```

src/dm.rs
agrover: 36 is what's in kernel 4.13 now, so that's what we need to support until kernel 4.14 comes out. The fix will also be backported to 4.13-stable.

src/dm.rs
agrover: `...` (three periods) indicates inclusive range, see https://doc.rust-lang.org/1.6.0/book/patterns.html#ranges

DM major version hasn't changed since at least 2005. If it ever changes (doubtful) we'll need to rewrite the whole crate, so checking it here doesn't add value.

src/dm.rs
agrover: oh. the word "now". ok 😃

agrover:
fixed comment.

src/dm.rs
mulkieran: Whoops! Why does search https://duckduckgo.com/?q=rust+syntax+Range+three+dots&t=ffsb&ia=qa give me Ruby documentation?

mulkieran: We can ignore major version as:

```
match hdr.version[1] {
  0...35 => None,
  minor => {
    // A bug in version <current major>.36 was corrected in subsequent versions.
    // <link to bugzilla or other>
    // This bug forces a different calculation of offset value.
    let offset = if minor == 36 {
       ...
    } else {
        12 + slc.len() + 1
    }
    
    let aligned_offset ...
    ...
  }
}
```

agrover: My goal with this code was to make it possible for us to use DM events in the short term, but after kernels supporting 37 are widespread there's no benefit to supporting 36. Delete the code and lump 36 in with the previous versions and return None. Enmeshing the 36 and 37 handling now is going to make removing the 36 code require more thought and create the risk of breaking 37 code in doing so. stet.

mulkieran: Aaah. Nobody told _me_ that. Why not just make the comment very definite, like:

```
    // A bug in minor version 36 was corrected in subsequent versions.
    // <link to bugzilla or other>
    // This block is a makeshift to allow early access to the event number value.
    // Once minor version 37 is widely available, this block is to be removed
    // and other necessary changes made so that event number is None for minor version 36.
```

agrover:
Updated comment to make it clear version 36 support is temporary.

mulkieran:
I read the thread and I think I may understand better what is going on. Here are my comments:

* I think that the type for the event number should actually be ```Result<Option<u32>>```, so that None can be used for the zero value, which I think means "no event number" in this context. devicemapper can provide a new error value indicating that that result is unsupported, like Unsupported(<some version info>) which means not supported until such and such a version. The client can unwrap the result, and respond to the error as appropriate.
If 0 is used as an actual event number (hopefully not), then I think the result type should be ```Result<u32>``` instead of ```Option<u32>```.

* I agree that dropping support for v36 access to this value as soon as possible makes sense. AFAICT, it should fail to work on machines with unusually short integers, so it is a bug right now, not just a creator of future bugs. But, I think that a restructuring is still indicated in order to minimize the surgery.

```
match hdr_version[1] {
  minor@0...36 => {
    if minor == 36 {
      // suggested comment to explain unfortunate situation
      // chunk of code for 36 here
      Ok(if nr == 0 { None } else {Some(nr)})
    } else {
      Err(Unsupported(37))
    }
  },
  _ => {
    // chunk of code for 37 and above here
    Ok(if nr == 0 { None } else { Some(nr) })
  }
}
```

agrover:
Every device has an event_nr even if it never has an event. list_devices will return 0 for event_nr for all devices until one of them has an event, in which case that device's event_nr would be 1 and the rest would remain 0. (And we then figure out which device(s) had a new event by comparing the previous output of list_devices with the latest, to see which devices have changed event_nrs.) So each device in kern 4.13+ will always have a value for event_nr (possibly 0). The use of Option is to allow for pre-4.13 kernels where device event_nrs are not obtainable via list_devices, and Option seemed like a straightforward way to represent that.

mulkieran:
I think that Result has only the most marginal utility over Option in this situation. But I still think that the restructuring shown below is more in line with the stated purpose of getting rid of v36 special code as soon as possible.

```
match hdr_version[1] {
  minor@0...36 => {
    if minor == 36 {
      // suggested comment to explain unfortunate situation
      // chunk of code for 36 here
      Ok(nr) # OR Some(nr) if using Option 
    } else {
      Err(Unsupported(37)) # OR None is using Option
    }
  },
  _ => {
    // chunk of code for 37 and above here
    Ok(nr) # OR Some(nr) if using Option
  }
}
```

agrover:
rebased. Restructured list_devices logic as suggested.

mulkieran:
The code looks fine to me.

PR #184, opened 2017-09-29
Extend a test with a slightly more interesting table

mulkieran:
Only one line is not enough for a LinearDev as its tables are, in general,
more than one line.


mulkieran:
Now testing w/out the fix. I should like this to fail.

mulkieran:
@agrover review?

agrover:
great! thanks so much! merging.

PR #187, opened 2017-10-11
device: Convert to/from dev_t and kdev_t

### LINKED ISSUE #147: Handling dev_t, and major and minor parts
Currently, `Device` has a u32 for major ID and u8 for minor ID. We want to use a u32 for minor as well, since Linux and device-mapper support minor numbers greater than 2^8.

Also, the u64 that Device::from() takes is ambiguous. See /usr/include/bits/sysmacros.h. Currently it's doing a zero-extended version of the legacy 16 bit dev_t representation, but passing a glibc `dev_t` will not be correct. e.g. passing the return value from metadata.st_rdev() into Device::from() is wrong. So that's the issue with userspace dev_t.

The devno embedded in DM ioctls we also have an issue. They are encoded differently (as described accurately in that sysmacros.h file) so constructing a Device from that field will also need to be done properly.

agrover:
We can't assume which bits in a libc::dev_t are major and which are minor.
This is especially true since the default is to use an intermingled
pattern that allows for >256 major and minor, but keeps the layout the same
as the old 16 bit dev_t if both values are less than 256.

Also properly convert to/from the kernel's dev_t encoding. Even though
DM's ioctl struct has a 64 bit field for this, it uses the kernel's
"huge" encoding, which is only 32 bits. Therefore it's possible to have
a valid Device that cannot be expressed to the kernel. But, this field
isn't currently ever set right now. If we ever want to specify DM devices
by maj:min then we'd need to use to_kdev_t().

fixes #147

### DIFF
--- a/src/device.rs
+++ b/src/device.rs
@@ -3,16 +3,17 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 use std::fmt;
+use libc::{dev_t, major, makedev, minor};
 
 /// A struct containing the device's major and minor numbers
 ///
-/// Also allows conversion to/from a single 64bit value.
+/// Also allows conversion to/from a single 64bit dev_t value.
 #[derive(Debug, PartialEq, Eq, Clone, Copy, Hash)]
 pub struct Device {
     /// Device major number
     pub major: u32,
     /// Device minor number
-    pub minor: u8,
+    pub minor: u32,
 }
 
 /// Display format is the device number in "<major>:<minor>" format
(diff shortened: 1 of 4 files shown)


src/dm.rs
```diff
@@ -761,7 +761,10 @@ impl DM {
                                       target_deps.count as usize)
             };
 
-            Ok(dev_slc.iter().map(|d| Device::from(*d)).collect())
+            Ok(dev_slc
+                   .iter()
```
mulkieran: I have this feeling that you could do ```into_iter()``` and thereby avoid the dererencing.

src/dm.rs
```diff
@@ -761,7 +761,10 @@ impl DM {
...
+                   .map(|d| Device::from_kdev_t(*d as u32))
```
mulkieran: We should have some justification for the safety of this downcast. Or maybe the above should be cast to a pointer to a u32 and that matters?

src/device.rs
```diff
@@ -25,14 +26,74 @@ impl fmt::Display for Device {
 impl From<u64> for Device {
```
mulkieran: We should be more specific about where this u64 is required to come from in order to make this work, in comments or in code. It is more or less explained in the test code.

src/dm.rs
agrover: Tried it but `into_iter()` and `iter()` on a slice appear to do the same thing and give us a ref only.

src/dm.rs
agrover: ok will do.

agrover:
rebase.
* add note explaining cast in `table_deps()`
* Use from_kdev_t in `DeviceInfo::device()`
* Use `dev_t` type alias instead of `u64`

src/deviceinfo.rs
```diff
@@ -51,7 +51,9 @@ impl DeviceInfo {
 
     /// The device's major and minor device numbers, as a Device.
     pub fn device(&self) -> Device {
-        self.hdr.dev.into()
+        // dm_ioctl is reserves 64 bits for device but kernel "huge" encoding
```
mulkieran: Typo.

agrover:
squashed. waiting for CI to complete.

PR #190, opened 2017-10-17
Transfer a test from stratisd to devicemapper

mulkieran:
The test is really more appropriate in devicemapper than stratisd, as it tests
only devicemapper functionality.

The minimal thinpool has to be made slightly larger, so that an xfs filesystem can be made on a thindevice in the pool.

### DIFF
--- a/src/thinpooldev.rs
+++ b/src/thinpooldev.rs
@@ -304,7 +304,7 @@ pub fn minimal_thinpool(dm: &DM, path: &Path) -> ThinPoolDev {
                                 &[Segment::new(dev, Sectors(0), MIN_RECOMMENDED_METADATA_SIZE)])
             .unwrap();
 
-    // 8 * MIN_DATA_BLOCK_SIZE should be enough to mount an xfs filesystem
+    // 512 * MIN_DATA_BLOCK_SIZE (32 MiB) should be enough for an xfs filesystem
     let data = LinearDev::setup(dm,
                                 DmName::new("data").expect("valid format"),
                                 None,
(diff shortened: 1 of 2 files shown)


mulkieran:
I expect this will pass under Jenkins once it is running the loop tests, so now depends on: #191 .

mulkieran:
Another possible is that a minimal thinpool is somehow too small when actually mounting a filesystem.

mulkieran:
If we increase the the thindev to the whole size of the thinpool data part, it may be large enough for xfs.

mulkieran:
mkfs.xfs succeeds w/ this error, ```mkfs.xfs: libxfs_device_zero write failed: Input/output error```, and then mounting fails.

src/thinpooldev.rs
```diff
@@ -304,12 +304,14 @@ pub fn minimal_thinpool(dm: &DM, path: &Path) -> ThinPoolDev {
                                 &[Segment::new(dev, Sectors(0), MIN_RECOMMENDED_METADATA_SIZE)])
             .unwrap();
 
-    let data =
-        LinearDev::setup(dm,
-                         DmName::new("data").expect("valid format"),
-                         None,
-                         &[Segment::new(dev, MIN_RECOMMENDED_METADATA_SIZE, MIN_DATA_BLOCK_SIZE)])
-                .unwrap();
+    // 8 * MIN_DATA_BLOCK_SIZE should be enough to mount an xfs filesystem
```
trgill: update the comment for : 512u64 * MIN_DATA_BLOCK_SIZE that is allocated?

mulkieran:
@agrover @trgill review?

mulkieran:
Whoops. Looking at Stratis PR, I realize we should probably be moving one more test.

mulkieran:
@agrover review? This really is ready to go.

agrover:
👍 , merging.

PR #193, opened 2017-10-20
Introduce error_chain error infrastructure

mulkieran:

### DIFF
--- a/src/result.rs
+++ b/src/result.rs
@@ -4,10 +4,10 @@
 
 use std::error::Error;
 use std::fmt;
-use std::io;
-use nix;
 
-///
+use super::errors;
+
+/// A very simple breakdown of outer layer errors.
 #[derive(Debug)]
 pub enum ErrorEnum {
     /// generic error code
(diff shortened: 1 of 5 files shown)


mulkieran:
I'm achieving here the limited goal of introducing error_chain into devicemapper core. I don't think any of this is really wrong and it seems it will always get benefits for this subset of errors.

mulkieran:
@agrover @trgill review?

mulkieran:
rebased.

mulkieran:
Depends on #194 , since that PR eliminates an error return in ```table_load```. This in turn means that all method in dm.rs that wrap ```do_ioctl``` calls, can fail only on a ```do_ioctl``` failure.

mulkieran:
Sort of depends on #196, as well. Since those ```unwrap/expect``` calls might have been actual error returns.

src/result.rs
```diff
@@ -65,17 +56,12 @@ impl fmt::Display for DmError {
...
         }
     }
 
     fn cause(&self) -> Option<&Error> {
-        match *self {
-            DmError::Dm(_, _) => None,
-            DmError::Io(ref err) => Some(err),
-            DmError::Nix(ref err) => Some(err),
-        }
+        None
```
agrover: If we're just going to return None for cause(), this is the same as the default impl and we can just omit it.

mulkieran: great!

PR #195, opened 2017-10-24
Event rearm support

agrover:


agrover:
OK, thanks. I've added some more text to the docstring too.

mulkieran:
I think we're still fine here, so I'll go ahead and merge.

PR #199, opened 2017-11-03
Have IoctlError include header struct field

mulkieran:
The purpose here is to have IoctlError include the header struct that was passed when the Ioctl called. It seems like the only obviously useful context.

In order to do that I had to implement Debug for ```Struct_dm_ioctl```. Then in passing, I noticed that ```Struct_dm_ioctl``` derived Copy, but was rather large for that, and that it wasn't really used, except to implement Clone, which I then implemented explicitly.


mulkieran:
@agrover @trgill review?

mulkieran:
This needs a few small fixes...

mulkieran:
clippy has a bug...the unused import it reports is not unused. I'll look into it more fully. Otherwise, this code seems to be passing, and I think it is ready to go.

@agrover @trgill review?

src/dm_ioctl.rs
```diff
@@ -94,6 +97,25 @@ impl ::std::default::Default for Struct_dm_ioctl {
...
+            .field("version", &self.version)
+            .field("data_size", &self.data_size)
+            .field("data_start", &self.data_start)
+            .field("target_count", &self.target_count)
+            .field("open_count", &self.open_count)
+            .field("flags", &self.flags)
+            .field("event_nr", &self.event_nr)
+            .field("padding", &self.padding)
+            .field("dev", &self.dev)
+            .field("name", &self.name.to_vec())
```
agrover: There might be a few ways to improve these three lines. First, there's an allocation -- not a huge deal but hey maybe we can avoid it. Second, printing a vec of `i8`s is going to be ugly. Third, it will always print the entire length, when the valid bytes stop at the first NULL.

To get these looking better, how about
 ```
.field("name", &String::from_utf8_lossy(&self.name))
```
? I think this will get the line printing as a string instead of a vec, and avoid the allocation. We may also need a call to `slice_to_null` in there somewhere, to resolve the third issue.

Note that the above line will not compile because the arrays are defined as libc::c_char, which is an `i8`. Maybe we should just change the array definitions for name, uuid, and data to `u8`? This could also simplify other parts of dm.rs, where we have to cast `i8` arrays to `u8`, which is pointless since we exclusively treat them as `u8`.

File: src/errors.rs [hunk 2]
```diff
@@ -24,7 +26,7 @@ error_chain! {
         /// An error returned exclusively by DM methods.
         /// This error is initiated in DM::do_ioctl and returned by
         /// numerous wrapper methods.
-        IoctlError {
+        IoctlError(t: Box<dmi::Struct_dm_ioctl>) {
```
agrover: Why a `Struct_dm_ioctl`? Why not a `DeviceInfo`? It's basically the same thing but with convenience methods for accessing fields in a more friendly manner. This would also let us keep the option to refactor DeviceInfo's internal representation, which we could never do if we return a raw `Struct_dm_ioctl`.

mulkieran: Done.

src/dm_ioctl.rs
mulkieran: Printing the entire vector was a deliberate choice. This is the implementation of Debug, and I think that really means that we ought to be looking at the whole structure, and not trying to make it look pretty. We really risk hiding the cause of the bug by doing anything else, I think.

mulkieran: Remember that we are passing the whole thing to some C code, and things cold happen to make the null termination be ignored. Kind of like here: https://bugzilla.redhat.com/show_bug.cgi?id=1312146.

agrover: I hear what you're saying about completeness but I also find looking at a string as a `Vec<u8>` pretty unpleasant. Can we print it as a string but include embedded NULLs, perhaps?

Also, what did you think about making the `c_char` arrays into `u8`s, either in this PR or a later one?

mulkieran: Can we make possibly prettifying the Debug output an issue rather than discussing it now? I absolutely agree that it is ugly, but we may still find that it is better than the alternatives.

mulkieran: I don't know enough about how the Struct definitions were generated to know what to think about the c_char question.

agrover: yup ok, let's defer prettifying and c_char questions.

PR #205, opened 2017-11-07
Generate bindings using rust-bindgen as part of build

agrover:
This changes from using the one-time output of rust-bindgen to generate
DM ioctl struct definitions, to generating it as part of the build process.
This keeps the definitions updated if the header file changes.

It's possible to customize generation as needed by configuring the builder
in build.rs. For example, traits for structs we wish to write ourselves can be omitted from generation. See [here](https://docs.rs/bindgen/0.31.3/bindgen/struct.Builder.html) for more.

Most of the changes are because the initial bindgen run added "Struct_" on struct names, whereas the current version doesn't.

The only tricky bit is that if one needs to see the generated code,
it is now are stuck deep in target/, in a file called bindings.rs. (This is following the example in the Bindgen tutorial, although I suppose we could put it in a different place if we wanted to.)


agrover:
see https://github.com/rust-lang-nursery/rust-bindgen/issues/1144

agrover:
inactive, closing.

PR #206, opened 2017-11-07
Get rid of frequently repeated comment

mulkieran:
It is fairly clear why empty flags are passed, so the comment is not
necessary.

### DIFF
--- a/src/dm.rs
+++ b/src/dm.rs
@@ -198,7 +198,6 @@ impl DM {
     pub fn version(&self) -> DmResult<(u32, u32, u32)> {
         let mut hdr: dmi::Struct_dm_ioctl = Default::default();
 
-        // No flags checked so don't pass any
         Self::initialize_hdr(&mut hdr, DmFlags::empty());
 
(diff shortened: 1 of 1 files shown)


mulkieran:
@agrover @trgill review?

Please take a look at the second commit. I haven't completed this step for all relevant methods in dm, but it's a good example. Do you have any concerns or suggestions at this time?

agrover:
Do we already get this info from the stack trace?

mulkieran:
@agrover Which info?

agrover:
Error chain appears to include backtrace, which gives us calling function. And IoctlError includes entire DeviceInfo. What more info does IoctlWrapperError give us? Also, suggest moving to its own PR since this is not related to the trivial comment commit.

mulkieran:
Due to the fact that the DM functions that call ```do_ioctl``` have been cleaned up to the point that they have, it doesn't give a whole lot of benefit. I'm willing to give up on this and consider error chaining of devicemapper core portion to be complete.

mulkieran:
Ok. Now this PR does what it says on the tin.

@agrover review?

PR #210, opened 2017-11-17
Verify that data block size is within bounds in ThinPoolDev methods

mulkieran:
It's not costly to do, and it uses some otherwise dead code.

It's a lot clearer and more specific than the ioctl return code which is
what would otherwise be returned on bad values.

Export the minimum and maximum values that we believe in.


mulkieran:
This is of dubious utility. We might be better off making a special type for data block size.

mulkieran:
This is the wrong way to go about it, see #246, but the PR does have some useful constants.

PR #213, opened 2017-11-17
Some work on messages

mulkieran:
Introduce a generic way of messaging a device. This should be just as useful for messaging cache devs as for messaging thinpool devs.


mulkieran:
@agrover @trgill review?

agrover:
👍

PR #215, opened 2017-11-21
Use macro_attr instead of custom_derive

mulkieran:
The use of custom_derive has been deprecated.

newtype_derive can no longer be used within macro_attr as with custom_derive.
Therefore, they are used independently.


mulkieran:
@agrover review?

PR #228, opened 2017-12-06
Fix bug when parsing a device with 3 or more tables

agrover:
targ.next (next_off) apparently counts from the start of the entire buffer,
not the end of the previous target entry! This only bites us when a device has 3
or more entries.


agrover:
@mulkieran review?

agrover:
Yeah it doesn't seem like our tests have the capability to load tables on DM devices. Looking fwd to your improvements.

Yes, three lines in a single table. Thanks.

PR #217, opened 2017-11-21
Disentangle outer devicemapper layer from core

mulkieran:
Should be a lot tidier.

Signed-off-by: mulhern <amulhern@redhat.com>

PR #216, opened 2017-11-21
Strip out all non-core parts of devicemapper

mulkieran:
There are compelling aesthetic and practical reasons for splitting
existing library into two parts.

Signed-off-by: mulhern <amulhern@redhat.com>

PR #238, opened 2017-12-18
Split devicemapper-rs into two crates using workspaces

mulkieran:


mulkieran:
Jenkins tests will not pass. They need to be allocated to the separate crates, somehow.

mulkieran:
@agrover @tasleson @trgill review? I don't like the extra complexity of using workspaces compared to just having separate repos and separate crates: see #217, #216. But it should still make for a better experience than keeping the two parts of devicemapper completely undifferentiated.

mulkieran:
rebased.

agrover:
inactive, closing for now.

Issue #230, opened 2017-12-11
Might be a good idea to use a dedicated type for each device's table


mulkieran:
A dedicated table type would still have to give a view of itself as a slice of TargetLine's or a slice of ```TargetLine<String>``` for benefit of ```DM::table_load()```. ```DmDevice::table()``` would return a ```T: DmTable```. So, DmTable would need a method to construct itself from a slice. DmTable would get ```equivalent_devices()``` method, getting rid of ```equivalent_tables()``` in ```DmDevice```. The only obvious benefit is that ```DmTable``` could do some checking of correctness while constructing itself from a slice in order to avoid checking of correctness in ```equivalent_devices()```. This seems kind of encouraging, actually.

mulkieran:
Resolved by #255.

Issue #229, opened 2017-12-11
Using ```dm_table``` when reloading a table is a latent bug


mulkieran:
The reason for this is that ```dm_table``` sets defaults. When a table is reloaded, we usually want to change only the thing that we intended to change. By using ```dm_table``` we will end up setting the values in the table back to their original defaults, whatever those may be.

mulkieran:
We need to fix this sooner rather than later, since this code is ending up in CacheDev as well.

mulkieran:
Resolved by #255.

Issue #220, opened 2017-11-28
Break into high-level and low-level layers


agrover:
@mulkieran I just wanted to open this since you're working on it (#216, #217) but there was not an issue (unless I missed it.) Do you want to talk about why you think this is a good idea? Are you aiming for separate crates, or separate git repos? (These need not be tied, see "cargo workspaces".) and why is this better than the status quo?

mulkieran:
Also #27.

mulkieran:
Broke the core stuff out into a sub-module a while ago, and it doesn't make sense to have further plans to break out into separate libraries at this time, so might as well close.

PR #239, opened 2017-12-20
Use a TargetTable trait to represent a device's table

mulkieran:
Resolves: #229.
Resolves: #230.
Resolves: #244.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,7 @@ macro-attr = "0.2.0"
 serde = "1"
 clippy = {version = "0", optional = true}
 error-chain = "0.11"
+either = "1"
 
 [dev-dependencies]
 loopdev = "0.2"
--- a/src/cachedev.rs
+++ b/src/cachedev.rs
@@ -16,6 +16,10 @@ use super::shared::{DmDevice, TargetLine, TargetParams, TargetTable, device_crea
                     device_match, parse_device};
 use super::types::{DataBlocks, DevId, DmName, DmUuid, MetaBlocks, Sectors, TargetTypeBuf};
 
+
+const CACHE_TARGET_NAME: &str = "cache";
+
+
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub struct CacheDevTargetParams {
     pub meta: Device,
@@ -154,7 +158,7 @@ impl FromStr for CacheDevTargetParams {
 
 impl TargetParams for CacheDevTargetParams {
     fn target_type(&self) -> TargetTypeBuf {
-        TargetTypeBuf::new("cache".into()).expect("< max length")
+        TargetTypeBuf::new(CACHE_TARGET_NAME.into()).expect("< max length")
     }
 }
 
@@ -218,7 +222,7 @@ impl TargetTable for CacheDevTargetTable {
         }
         let line = table.first().expect("table.len() == 1");
         let target_type = &line.2.to_string();
-        if target_type != "cache" {
+        if target_type != CACHE_TARGET_NAME {
             let err_msg = format!("Parsing a cache table entry but found target type {}",
                                   target_type);
             return Err(DmError::Dm(ErrorEnum::Invalid, err_msg));
--- a/src/cachedev.rs
+++ b/src/cachedev.rs
@@ -160,6 +160,22 @@ pub struct CacheDevTargetTable {
     table: TargetLine<CacheDevTargetParams>,
 }
 
+impl CacheDevTargetTable {
+    pub fn new(start: Sectors,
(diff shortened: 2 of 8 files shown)


mulkieran:
Rebased roughly.

mulkieran:
rebased.

mulkieran:
rebased.

mulkieran:
rebases and changes.

mulkieran:
@trgill Hi Jenkins test failed, but I can't look at it, and it seems to prevent Travis from running.

trgill:
@mulkieran OpenStack nodes are not responding to Jenkins.  Looks like all? of them... Will track it down.

mulkieran:
@trgill Thanks!

mulkieran:
spurious Jenkins failure.

mulkieran:
Get rid of dead code annotation.

mulkieran:
@agrover This builds on previous reorganization work and makes it possible to have a LinearDev with flakey bits. It has not been completed, looking for any immediate feedback you have.

mulkieran:
Probably the right thing to do is to extend Segment with an optional field for flakey info, a struct that has
```up_interval```, ```down_interval```, and feature args.  If it is not ```None``` then target is flakey. This looks straightforward, I think the only function that would need to be altered is ```LinearDevTargetTable::new```.

agrover:
The LinearDev doesn't know the type of the blockdevs it is mapping. They could be any blockdev, either DM or non-DM. Their params (if any) aren't visible to the LinearDev. `LinearDevTargetParams` seems to be constraining each segment making up a Linear to be either another Linear or a Flakey, which doesn't seem right.

mulkieran:
@agrover I don't think that's true. These targets just constrain how the linear dev views the underlying device, not what the underlying device is. So, a linear target can map over any type of underlying device. A flakey target can do the same.

There is another point, which is that currently each line in a linear dev can only be flakey or linear target, due to the definition of LinearDevTargetParams. I think there is also a delay target, which could be added later without any great difficulty, and just another use of Either constructor.

mulkieran:
There is a tweak that I think will make this a bit cleaner. Have TargetParams include the ```target_type``` value. In that case, parseing and display will include this value, and even LinearDevTargetParams can implement FromStr by reading the first value, which will identify the target type and then dispatching to the appropriate TargetParam implementation, either Flakey or Linear.

mulkieran:
@agrover review?

agrover:
After digging around in the kernel code, the list of DM devs that can only have a single target (.features include DM_TARGET_SINGLETON) is:
* thin-pool
* mpath
* integrity
* zoned

AFAICT *all* the other ones (10+) can be used for devices that have multiple target lines in their table. (Doing so might be silly in some cases, but it's possible).

So yeah, there is no such thing as a FlakeyDev, only a device with a "flakey" target line (among possibly others). No such thing as a LinearDev.

Good to get clear on all this, but this suggests we will need to rework our dm upper layer at some point, to not artificially restrict the possible DM configurations.

mulkieran:
The restriction you've found seems incomplete to me.It's easy for me to understand why a thin-pool target must be the unique one on a DM device, but it is hard for me to understand why that same restriction wouldn't apply for a cache target. Maybe someone just forgot to set ```DM_TARGET_SINGLETON``` for some targets, or didn't want to go to the trouble.

Anyway, this PR does go in  the right direction, by setting up LinearDev so that it can use either type of target, flakey or linear in a way that makes sense. Running together a bunch of cache and thindev targets on a single device does seem a little silly to me. So, it probably won't hurt us to leave all that alone for now.

After all, so long as the upper layer isn't actually wrong, it's probably alright if it reflects our prejudices.

This is one of the many reasons why I think the upper layer should end up in a separate library. I can't imagine we'ld ever have these sorts of discussions about the core part of the library. It's a completely different kettle of fish.

agrover:
>This is one of the many reasons why I think the upper layer should end up in a separate library. I can't imagine we'ld ever have these sorts of discussions about the core part of the library. It's a completely different kettle of fish.

I opened this so we can discuss independently of other ongoing PRs: https://github.com/stratis-storage/devicemapper-rs/issues/220

agrover:
> The restriction you've found seems incomplete to me.It's easy for me to understand why a thin-pool target must be the unique one on a DM device, but it is hard for me to understand why that same restriction wouldn't apply for a cache target. Maybe someone just forgot to set DM_TARGET_SINGLETON for some targets, or didn't want to go to the trouble.

I don't know, maybe we can ask?

mulkieran:
Look at #255 instead.

mulkieran:
Superceded by #255.

PR #241, opened 2018-01-02
Do not use TargetLine in devicemapper core methods

mulkieran:
This takes the use of TargetLine and TargetParams, in terms of which TargetLine is defined out of core methods and confines its use to the outer layer.

TargetLine can only be used in one place in the core, ```DM::table_load```, and the only benefit to this use is using names instead of indices to access values. One could go further, and parameterize ```DM::table_load``` on the TargetParams, but then you get as many copies of this method as there are types of device, thin, thinpool, etc., all for the benefit of calling ```to_string()``` in exactly one place in the method. I'm pretty sure that is not enough benefit.

### DIFF
--- a/src/cachedev.rs
+++ b/src/cachedev.rs
@@ -12,9 +12,9 @@ use super::deviceinfo::DeviceInfo;
 use super::dm::{DM, DmFlags};
 use super::lineardev::LinearDev;
 use super::result::{DmResult, DmError, ErrorEnum};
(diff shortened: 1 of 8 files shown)


mulkieran:
Cherry-picked from #239.

mulkieran:
@agrover @trgill review?

agrover:
👍 , MALR

Issue #232, opened 2017-12-11
Refine equivalent_tables() methods to determine equivalence of devices


mulkieran:
Right now, all but cache dev use simple equality on the structs. This is less likely to result in false negatives than string comparison, but there will still be some. These will be fairly unusual, though, as tables won't change much from their defaults.

PR #243, opened 2018-01-02
Be more explicit about default values in tables

mulkieran:
This changes the name of internal ```dm_table``` function, and implements ```equivalent_tables``` for all devices.

Resolves: #232.

### DIFF
--- a/src/shared.rs
+++ b/src/shared.rs
@@ -44,9 +44,7 @@ pub trait DmDevice<T: TargetParams> {
     fn device(&self) -> Device;
 
     /// Check if tables indicate an equivalent device.
-    fn equivalent_tables(left: &[TargetLine<T>], right: &[TargetLine<T>]) -> DmResult<bool> {
-        Ok(left == right)
-    }
+    fn equivalent_tables(left: &[TargetLine<T>], right: &[TargetLine<T>]) -> DmResult<bool>;
 
     /// The device's name.
     fn name(&self) -> &DmName;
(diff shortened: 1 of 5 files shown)


mulkieran:
@agrover @trgill review please?

mulkieran:
Cherry-picked from #239.

src/shared.rs
```diff
@@ -23,9 +23,7 @@ pub trait DmDevice<T: TargetParams> {
     fn device(&self) -> Device;
 
     /// Check if tables indicate an equivalent device.
-    fn equivalent_tables(left: &[TargetLine<T>], right: &[TargetLine<T>]) -> DmResult<bool> {
-        Ok(left == right)
-    }
+    fn equivalent_tables(left: &[TargetLine<T>], right: &[TargetLine<T>]) -> DmResult<bool>;
```
agrover: Only comment is: will it let you define a default implementation of `Ok(left == right)` here, that each implementer can use, since they're all the same? Other than that, looks great, much clearer.

mulkieran: I think that it is likely that pretty soon that implementation won't be shared any more, so it is better to leave as is.

mulkieran:
rebased.

mulkieran:
@trgill review?

src/shared.rs
agrover: ok, makes sense.

agrover:
👍

PR #250, opened 2018-01-05
Make a special make target for fmt test on Travis

mulkieran:
Use rustup to set the version of rustfmt so that it is recognized.
Modify the existing fmt target so that it just reformats text inline.
This seems the handier thing for regular command-line users.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,7 +13,7 @@ cache: cargo
 matrix:
     include:
         - rust: stable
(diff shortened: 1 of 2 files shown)


mulkieran:
I'm calling this trivial, because I did the same thing on stratisd and it's already been approved and merged. So, if this fixes the problem, I'll just go ahead and merge it.

PR #251, opened 2018-01-05
add flakeydev target

trgill:


src/flakeydev.rs
```diff
@@ -0,0 +1,345 @@
...
+use super::deviceinfo::DeviceInfo;
+use super::dm::{DM, DmFlags};
+use super::lineardev::LinearDev;
+use super::result::{DmResult, DmError, ErrorEnum};
+use super::shared::{DmDevice, device_create, device_exists, parse_device, TargetLine, TargetParams};
+use super::types::{DevId, DmName, DmUuid, Sectors, TargetTypeBuf};
+
+
+#[derive(Debug, Eq, PartialEq)]
+pub struct FlakeyDevTargetParams {
```
mulkieran: Unusually, it makes sense to make _all_ fields public. The reason is that ```table()``` will return these as part of TargetLine and client would very probably want to know these values.

src/flakeydev.rs
```diff
@@ -0,0 +1,345 @@
...
+impl DmDevice<FlakeyDevTargetParams> for FlakeyDev {
+    fn device(&self) -> Device {
+        device!(self)
+    }
+
+    fn devnode(&self) -> PathBuf {
+        devnode!(self)
+    }
+
+    fn equivalent_tables(left: &[TargetLine<FlakeyDevTargetParams>],
```
mulkieran: Should get a comment, like thinpooldev's, indicating that the method is probably incomplete.

src/flakeydev.rs
```diff
@@ -0,0 +1,345 @@
...
+    fn uuid(&self) -> Option<&DmUuid> {
+        uuid!(self)
+    }
+}
+
+
+impl FlakeyDev {
+    /// new
+    pub fn new(dm: &DM,
+               name: &DmName,
```
mulkieran: should get optional DmUuid arg, like others.

src/flakeydev.rs
```diff
@@ -0,0 +1,345 @@
...
+               up_interval: u64,
+               down_interval: u64,
+               feature_args: Vec<String>)
+               -> DmResult<FlakeyDev> {
+        if device_exists(dm, name)? {
+            let err_msg = format!("flakeydev {} already exists", name);
+            return Err(DmError::Dm(ErrorEnum::Invalid, err_msg));
+        }
+
+        let table = FlakeyDev::dm_table(&linear_dev,
```
mulkieran: ```dm_table``` should be renamed to ```gen_default_table```.

src/flakeydev.rs
```diff
@@ -0,0 +1,345 @@
...
+                                        offset,
+                                        up_interval,
+                                        down_interval,
+                                        feature_args);
+        let dev_info = device_create(dm, name, None, &table)?;
```
mulkieran: should pass uuid parameter here, instead of ```None```.

src/flakeydev.rs
```diff
@@ -0,0 +1,345 @@
...
+
+        Ok(FlakeyDev {
+               dev_info: Box::new(dev_info),
+               linear_dev: linear_dev,
+               start_offset: offset,
+           })
+    }
+    /// setup
+    pub fn setup(dm: &DM,
+                 name: &DmName,
```
mulkieran: optional uuid here, as w/ ```new()```.

src/flakeydev.rs
```diff
@@ -0,0 +1,345 @@
...
+                 feature_args: Vec<String>)
+                 -> DmResult<FlakeyDev> {
+
+        let table = FlakeyDev::dm_table(&linear_dev,
+                                        offset,
+                                        up_interval,
+                                        down_interval,
+                                        feature_args);
+        let dev = if device_exists(dm, name)? {
+            return Err(DmError::Dm(ErrorEnum::Invalid, "device already exists".into()));
```
mulkieran: Why is error if device exists correct action? If it is correct action, maybe all this impl needs is a ```new()``` method?

src/flakeydev.rs
```diff
@@ -0,0 +1,345 @@
...
+
+    use super::super::consts::{IEC, SECTOR_SIZE};
+    use super::super::device::{Device, devnode_to_devno};
+    use super::super::loopbacked::{test_with_spec, write_sectors};
+    use super::super::segment::Segment;
+    use super::super::types::Bytes;
+
+    use super::*;
+
+    fn test_flakey_dev(paths: &[&Path]) -> () {
```
mulkieran: Needs a header comment. What did you do to make sure that writing sectors would fail?

mulkieran:
superceded by #255.

PR #252, opened 2018-01-12
Suspend device before loading table

mulkieran:
This is intuitively the correct thing to do.

### DIFF
--- a/src/shared.rs
+++ b/src/shared.rs
@@ -145,8 +145,8 @@ pub fn table_reload<T: TargetParams>(dm: &DM,
         .iter()
         .map(|x| (x.start, x.length, x.target_type.clone(), x.params.to_string()))
         .collect::<Vec<_>>();
-    let dev_info = dm.table_load(id, &table)?;
     dm.device_suspend(id, DmFlags::DM_SUSPEND)?;
+    let dev_info = dm.table_load(id, &table)?;
     dm.device_suspend(id, DmFlags::empty())?;
     Ok(dev_info)
 }


mulkieran:
@agrover review?

mulkieran:
I think that this is an obvious fix to a bug, but I believe that it turns out that we have bigger problems and that those mean that we should not be including suspend and resume calls in ```table_reload``` at all. We need to manage suspend and resume at a higher level. For example, when we set the segments of the data device of a thin pool, it seems that it is necessary to:
1. Suspend the thinpool.
2. Suspend the data device
3. Reload the data device table.
4. Resume the data device.
5. Reload the thin pool device table.
6. Resume the thin pool device.

What we have here is a nested structure, and our current setup breaks that. Do we know if suspending and resuming are idempotent? That would make things easier.

agrover:
So the way this works is that `table_load` does not replace the active table, it loads it into the "inactive" slot, and then whatever is in the inactive slot is moved to the "active" slot when the device is suspended/resumed. Based on the comments in dm-ioctl.h, looking at (table_load, suspend, resume) sequence, table_load loads the new (inactive) table, suspend flushes all pending i/o, and resume makes the inactive table active. Doing the suspend before the table_load seems allowed, but keeps the device suspended marginally longer than doing the table load first. Note that it is possible to resume a non-suspended device to make the inactive table active without flushing i/o. This seems like something one might need to do in some edge case, but unwise for the common case.

Also, suspend and resume are indeed idempotent.

mulkieran:
Thanks for the explanation!

It is good that suspend and resume are idempotent, because that makes things slightly easier for us.

Right now our code for setting the thinpool data segments:
1. Loads a new table for the data device into the inactive slot.
2. Suspends the data device.
3. Resumes the data device.
4. Loads a new table for the thin pool device into the inactive slot.
5. Suspends the thinpool device.
6. Resumes the thinpool device.

But the test code in ```devicem-mapper-test-suite``` follows the order in .

agrover:
Seems best to be consistent, even if both sequences are not "wrong" per se.

mulkieran:
I'm dropping this in favor of issue #253.

PR #255, opened 2018-01-23
Use the last table set as the essential attribute of the device

mulkieran:
Expected to supercede #239.


mulkieran:
There is a theory that says we would all be better off if client code build the table for LinearDevs. The Segments would not exist, and BldDevSegments in stratisd would map UUIDs to TargetLines.

mulkieran:
There is a flaw. ```load_table()``` returns an error if the table that it is attempted to load is a valid table but does not satisfy our constraints. If we doen't change that in this PR, it should be left as an issue. But it is no problem right now.

mulkieran:
The other flaw is that, although flakey targets are expressible in LinearDev, there is no way to actually construct them. But, I think that is part of our bigger flaw, and that Segments is not necessarily what we want.

mulkieran:
@agrover review?

agrover:
Hi, some comments/questions.

* 84fb6ef  "This makes it a whole lot more generally useful. " how?
* consider making LinearDevTargetTable and friends a `type` rather than a struct. Or a [tuple struct with single member](https://doc.rust-lang.org/1.6.0/book/structs.html#tuple-structs)? (Seeing self.table.table in the code tweaked my antennae.)
* 6432bd1 "It takes away the default organization of
the segments passed, which is contiguous, and leaves the responsibilty
for that arrangement up to the client. " Is there ever a use case for gaps in target lines? I can't think of one. Not sure if DM allows even. Is it possible to take "start" out of TargetLine, just leaving length and params, and only allow the building of contiguous tables? Less for callers to get wrong?

mulkieran:
> * 84fb6ef  "This makes it a whole lot more generally useful. " how?

I've edited the commit message.

mulkieran:
> consider making LinearDevTargetTable and friends a type rather than a struct. Or a tuple struct with single member? (Seeing self.table.table in the code was what made me go "huh?".

The reason for making this a struct rather than a type is that it defines an implementation for xforming from DM core raw tables to more informative outer-layer tables.

A tuple struct would be suprisingly inconvenient, because of the necessity to destructure and then reconstruct the tuple in the same method. It is a little wierd, in places, but I think it should be left as is.

mulkieran:
Two experimental additional commits.

mulkieran:
Last commit is really of uncertain value, but does make for less code.

mulkieran:
@trgill review?

mulkieran:
Note that there is a FIXME about handling FlakeyDev features args, since the don't follow the same format as those in cachedev and thinpooldev. If this gets merged, I'll open an issue. However, I don't think it is urgent.

trgill:
@mulkieran the CI failures look real, right?

mulkieran:
@trgill Yes, it's real. Oh, why did I add that test!???

mulkieran:
That the test is failing is very hard to explain. We have every reason to believe that the proper params are being uploaded in the table, because when we parse them after creating the device, we get what we would expect.

Is it possible that up an down interval are reversed in the docs? That would be crazy.

mulkieran:
Maybe FlakeyDev only works if the dev in its params is itself a linear dev? This is _not_ what the docs say, but it seems to be the only real difference between original test in flakeydev PR and this PR.

mulkieran:
New version of clippy is not available yet...probably in a few hours.

mulkieran:
We need an explanation for some observations:

1. First test
One lineardev with a single flakey target referring to a loop device. Configuration parameter are up_interval: 0, down_interval: 1. This should have an always down sort of effect. Result: Bits are written successfully to the linear device. Surprise!

agrover:
Yeah, I looked into the dm-flakey kernel code a little but it wasn't immediately obviously buggy. I agree with @mulkieran that we can tackle getting to that bottom of those unexplained behaviors in a separate PR.

So I am 👍 on this PR up through 32640cb .

mulkieran:
Removed test stuff at tip of PR.

mulkieran:
We will want to merge this https://github.com/stratis-storage/stratisd/pull/700 once new devicemapper release is out.

PR #259, opened 2018-01-31
Derive fmt::Display for newtypes instead of implementing ourselves

agrover:
Display is used by types implementing std::string::ToString trait. Our
custom implementations of Display add the units as a word, which seems to
be overly assuming in how Display is used.

Instead, use NewtypeDisplay, which just delegates to the wrapped type.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -148,6 +148,7 @@ NewtypeDeref! { () pub struct DataBlocks(u64); }
 NewtypeFrom! { () pub struct DataBlocks(u64); }
 NewtypeSub! { () pub struct DataBlocks(u64); }
 NewtypeSubAssign! { () pub struct DataBlocks(u64); }
+NewtypeDisplay! { () pub struct DataBlocks(u64); }
 
 self_div!(DataBlocks);
 serde!(DataBlocks);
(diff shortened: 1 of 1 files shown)


agrover:
@mulkieran as you know we're returning strings for properties such as TotalPhysicalSize. When this changes, we should be sending out a PropertyChanged signal of e.g. "34234234" not "34234234 sectors".

What was the reason for originally including the units in Display? It strikes me as non idiomatic.

mulkieran:
We must be sending out the value as a bare number currently, otherwise stratis-cli would struggle to deal with it. However we achieve that when we first set the property, we should do the same when the property changes.

agrover:
Yes we are dereffing:
```
get_pool_property(i, p, |p| Ok(format!("{}", *p.total_physical_size())))
```
But wouldn't it be better if this was:
```
get_pool_property(i, p, |p| Ok(p.total_physical_size().to_string()))
```
?

mulkieran:
We have been implementing Display for these types in this way for exactly a year, starting with a stratisd commit.
We use it under the expectation that it works that way in many places. See:

```
commit 4a4ec250830fd9bd230a0feefc5633a904b570a1
Author: mulhern <amulhern@redhat.com>
Date:   Tue Jan 31 10:19:15 2017 -0500
```

agrover:
I'd really prefer it work the other way. I'm happy to adjust spots affected in stratisd by this change.

mulkieran:
Before you change the code, it would make sense to come up with a compelling commit message explaining why you'ld much rather do that than add a ```*``` in one single place, which is what we have been doing when we have a Sector and need to display the bare number, for a whole year.

I do not see anything unidiomatic about the current setup. Display is supposed to be human readable, not for machine use, say the docs. I believe that anything you return on the D-Bus is for machine use.

mulkieran:
Another thing occurred to me, which is that there is not much difference between:

```get_pool_property(i, p, |p| Ok(p.total_physical_size().to_string()))```

and

```get_pool_property(i, p, |p| Ok((*p.total_physical_size()).to_string()))```

which ought to achieve what is wanted just as well as that ```format!``` expression.

agrover:
on hold, closing.

PR #262, opened 2018-02-08
Misc improvements

tasleson:


src/cachedev.rs
```diff
@@ -360,14 +360,14 @@ pub struct CacheDev {
...
 impl DmDevice<CacheDevTargetTable> for CacheDev {
-    fn device(&self) -> Device {
-        device!(self)
-    }
-
     fn devnode(&self) -> PathBuf {
         devnode!(self)
     }
 
+    fn device(&self) -> Device {
```
mulkieran: Please reorder the trait instead, so that ```device()``` is above ```devnode()```.

tasleson: Done!

Issue #263, opened 2018-02-08
nix 0.10


ignatenkobrain:
Would be nice if we could switch to nix 0.10

ignatenkobrain:
Same for stratisd

mulkieran:
This won't be as straightforward as I was hoping. nix 0.10 requires libc 0.2.36, but it appears that there is a least some other library that we use that transitively depends on libc that requires a lower version. We have to work out the puzzle.

```
[mulhern@localhost devicemapper-rs]$ cargo tree
devicemapper v0.16.0 (file:///home/mulhern/my-contributions/devicemapper-rs)
├── bitflags v1.0.1
├── error-chain v0.11.0
│   └── backtrace v0.3.4
│       ├── backtrace-sys v0.1.16
│       │   └── libc v0.2.34
│       ├── cfg-if v0.1.2
│       ├── libc v0.2.34 (*)
│       └── rustc-demangle v0.1.5
├── libc v0.2.34 (*)
├── macro-attr v0.2.0
├── newtype_derive v0.1.6
├── nix v0.9.0
│   ├── bitflags v0.9.1
│   ├── cfg-if v0.1.2 (*)
│   ├── libc v0.2.34 (*)
│   └── void v1.0.2
└── serde v1.0.24
```

mulkieran:
Whoops. I think I may just have gotten out of the mess.

agrover:
@ignatenkobrain I think this one is on its way to getting resolved but please next time if you give us a little bit more justification that'd be appreciated too 😃 Does this have something to do with your upstream nix work on making statvfs available on more archs?

ignatenkobrain:
@agrover sure! No, that's because I don't want to keep multiple versions of crates in Fedora :P

PR #264, opened 2018-02-09
Update nix to v0.10.0

mulkieran:
This requires an explicitly higher version of backtrace and of libc.

Resolves: #263.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,8 +11,8 @@ keywords = ["Linux", "device", "mapper", "libdm", "storage"]
 license = "MPL-2.0"
 
 [dependencies]
-libc = "0.2.31"
-nix = "0.9"
+libc = "0.2.36"
+nix = "0.10"
 bitflags = "1"
 newtype_derive = "0.1"
 macro-attr = "0.2.0"
--- a/src/device.rs
+++ b/src/device.rs
@@ -9,7 +9,7 @@ use std::path::Path;
 use std::str::FromStr;
 
 use libc::{dev_t, major, makedev, minor};
-use nix::sys::stat::{S_IFBLK, S_IFMT};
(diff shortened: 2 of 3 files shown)


mulkieran:
We shouldn't merge this until we can make sure we can update in stratisd as well. So there will be another PR forthcoming.

mulkieran:
@trgill We should expect to be able to merge this w/out passing Jenkins because I think that if we get this to pass all other Jenkins executions will fail. But, once this is merged, we'll need to update Jenkins pronto.

trgill:
@mulkieran there are only 3 Jenkins nodes.  I can update them as soon as we merge this...

Cargo.toml
```diff
@@ -11,8 +11,9 @@ keywords = ["Linux", "device", "mapper", "libdm", "storage"]
 license = "MPL-2.0"
 
 [dependencies]
-libc = "0.2.31"
-nix = "0.9"
+backtrace = "0.3.5"
```
ignatenkobrain: hmmm, why is this needed?

mulkieran: It may be unnecessary...I'm trying without.

mulkieran:
@trgill I may have overcomplicated the whole thing. When you do update the Jenkins nodes, maybe all you'll need to do is ```cargo update --package nix``` and the problems that I was encountering will be resolved without a hitch.

mulkieran:
Doesn't seem to require any code changes to stratisd. In order to not have Jenkins disabled for too long, we should delay merging this until we're about to do a new devicemapper release, I think.

agrover:
I got Jenkins passing by adding 'git clean -fxd' before calling 'make sudo_test'. @trgill does this seem reasonable?

PR #267, opened 2018-02-20
Document events

agrover:
Document how to poll for DM events.

Also, take the opportunity to remove support for the initial buggy `list_devices` layout.


agrover:
This is a feature of Markdown. If you render the docs with `cargo doc --open` you can see the numbers in an ordered list are made sequential.

However, this feature appears very under-documented, so not relying on it is probably best.

PR #276, opened 2018-02-27
Pass arguments to CacheDev::gen_default_table in proper order

mulkieran:
This is a bug fix. cache and origin order were swapped in CacheDev::setup.


mulkieran:
New version of clippy found new things to object to: #277.

agrover:
If clippy is the only thing blocking this then let's go ahead and merge. Don't care about clippy right now.

PR #277, opened 2018-02-27
Allow some clippy errors caused by bitflags! macro

mulkieran:
Hopefully bitflags maintainers will address these.


mulkieran:
@agrover review?

AFAICT, the test is compiling against the newest version of bitflags, so there isn't really any option of updating to a newer version.

Totally wierd: if you look at the Travis clippy target you'll notice that bitflags appears twice with different versions, but I think that the lower one is required by clippy. Given the dependency version that we specify in Cargo.toml, nothing else makes sense.

agrover:
don't care about clippy atm, esp. other peoples' clippy.

PR #278, opened 2018-02-27
Export META_BLOCK_SIZE in public interface of devicemapper

mulkieran:
It's a useful thing to compare various status values against.


mulkieran:
You can get this value by using existing functions, which is kind of safer.

PR #279, opened 2018-02-27
Add suspend() and resume() methods to DmDevice

mulkieran:
Rewrite methods that set tables to use these methods in what seems like the appropriate manner.

PR #282, opened 2018-02-28
Master public table

mulkieran:


mulkieran:
@trgill , review please?

Cherry-picked from #279.

src/cachedev.rs
```diff
@@ -172,7 +172,7 @@ impl TargetParams for CacheTargetParams {
 
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub struct CacheDevTargetTable {
-    table: TargetLine<CacheTargetParams>,
+    pub table: TargetLine<CacheTargetParams>,
 }
```
trgill: would it be better to add a method that returns the clone?

mulkieran: I took a look at the idea, but I don't think it would be useful. I think there will turn out to instances when the user just wants to look at the table, rather than change it in some way, and so they won't even want to clone. So, really what you want is a ```value()``` method on each implementation of TargetTable. You can get that in two ways. You can parameterizie TargetTable on the type of its value, but that is very intrusive. The other thing is to define methods on each of the TargetTable implementations, called, e.g. ```table()``` that get a reference to their internal table. But that's so very close to accessing the table field directly, that it just seems like overkill.

We don't have to worry about encapsulation, because if don't clone, if we try to do something to the table directly, we get a "cannot move out of borrowed content" error.

So, I think it's safer to stick with what we have here.

trgill: Ok.

PR #283, opened 2018-03-01
Rewrite set_*_table methods

mulkieran:


mulkieran:
@trgill review, please?

mulkieran:
In an excess of caution, I"m going to point my stratis branch at this before it gets committed.

mulkieran:
Whoops, these changes to a little bit too much.

mulkieran:
Seems like it's looking good again.

mulkieran:
Spurious fmt task failure.

mulkieran:
@agrover @trgill review again, please?

src/cachedev.rs
```diff
@@ -499,17 +500,25 @@ impl CacheDev {
                             dm: &DM,
                             table: Vec<TargetLine<LinearDevTargetParams>>)
                             -> DmResult<()> {
+        // Follow resize_origin example in device-mapper-test-suite file
+        // cache_stack.rb. This looks funky, but it does seem to be exactly
+        // what the test is doing.
```
agrover: No, we should resume self before returning from this fn to match that example. The `@cache.pause do` usage in that code calls resume when the do block is exited, see implementation of pause in dev.rb line 44ish and _bracket defined in prelude.rb. Clever, Joe!

In our code, it looks like most/all callers are calling `resume()` after these fns, so moving the resume inside is functionally the same, just a lil nicer.

mulkieran: We do not want to resume self. See commit message in "Do not resume" commit and header comments for these methods.

agrover: > If the top-level device requires a number of updates to its devices,
resuming in the middle of these updates could be harmful.

ok so it started out the other way :smiley:. I read the commitlog and comments but can you explain a little further the problems it caused? Does stratisd exhibit the behavior the commitlog alludes to?

mulkieran: @agrover I don't believe I succesfully made it exhibit the behavior. But when we are in a context where we must load in sequence new tables for all components of the thinpool device in the thinpool, i.e., the meta device and the data device, I am wary of resuming the thin pool with one sub-device still referring to the origin device while the other refers to the newly created cache device.

mulkieran: The comments about examples followed are not really useful anymore, though. I can remove them.

agrover: that makes sense, thanks.

mulkieran:
comments removed. Comments referring to devicemapper tests removed. They became misleading due to code changes.

PR #287, opened 2018-03-15
Move DmFlags definition into separate module

mulkieran:


mulkieran:
clippy failure.

mulkieran:
@agrover review, please?

mulkieran:
clippy should pass this time.

mulkieran:
And it does! @agrover review, please?

src/dm.rs
```diff
@@ -6,8 +6,6 @@
 // Their scope is the whole file because the bitflags! macro generates a
 // lot of stuff, and so placing the annotation right on the use of the
 // bitflags! macro does not actually cover the suspicious code generated.
-#![allow(redundant_field_names)]
```
agrover: the preceding comment is copied in the new file, can it be removed from here?

src/dm_flags.rs
```diff
@@ -0,0 +1,57 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+// These errors originate in bitflags! macro and are not ours to fix.
```
agrover: Are these clippy errors or regular rustc errors? Just curious. Might be nice to say.

agrover: Still an issue with most recent bitflags crate?

mulkieran: The most recent bitflags crate hasn't changed since these clippy errors were added.

mulkieran: These must be clippy errors, because we use clippy annotations to ignore them.

src/dm.rs
mulkieran: Yes.

mulkieran:
@agrover review? Turns out clippy change means those two lints aren't triggered by bitflags! generated code, so the whole clippy thing can go away.

PR #293, opened 2018-03-29
Add method to suspend a DM device with DM_NOFLUSH

trgill:
The DM_NOFLUSH option is required when expanding the meta or data devices for a thin-pool.


agrover:
:+1:

PR #295, opened 2018-03-30
Generate UML-diagram like graphs for devicemapper-rs

mulkieran:


mulkieran:
Jenkins uses sudo_test, so it doesn't like test-sudo. That's probably an easy change, though.

agrover:
Looks fine, but can we stick uml_graphs.rs in a directory? My first thought was docs/utils or docs/scripts or something...

mulkieran:
How about this? The subdirectory is just "scripts", which kind of makes sense for something that is run using ```cargo script```.

mulkieran:
Might as well consider it ready for final review.

PR #296, opened 2018-04-19
Set event_nr in IOCTL hdr

tasleson:
From guidance given to us by Peter Rajnoha, we need to set the upper 16 bits
of event_nr to 0x0040 (DM_UDEV_PRIMARY_SOURCE_FLAG) to get some good default
behavior from the dm-* udev rule files.  This and the information stated in
lvm2 commit 942d6ef29ff0e8ae5f76ca50ed7570ae8f7c5a66, it appears that this
flag should always be set.

Testing shows that we are getting udev db entries which allow us to be usable
with this flag set.

### DIFF
--- a/src/dm.rs
+++ b/src/dm.rs
@@ -30,6 +30,10 @@ const DM_VERSION_MINOR: u32 = 30;
 /// Patch level
 const DM_VERSION_PATCHLEVEL: u32 = 0;
 
+// See https://sourceware.org/git/?p=lvm2.git;a=blob;f=libdm/libdevmapper.h#l3708
+// Note: flag is only upper 16 bits.
+const DM_UDEV_PRIMARY_SOURCE_FLAG: u16 = 0x0040;
+
 /// Start with a large buffer to make BUFFER_FULL rare. Libdm does this too.
 const MIN_BUF_SIZE: usize = 16 * 1024;
 
@@ -58,6 +62,7 @@ impl DM {
         hdr.version[2] = DM_VERSION_PATCHLEVEL;
 
         hdr.flags = flags.bits();
+        hdr.event_nr = (DM_UDEV_PRIMARY_SOURCE_FLAG as u32) << 16;
 
         hdr.data_start = size_of::<dmi::Struct_dm_ioctl>() as u32;
     }


mulkieran:
I'ld like to put this on the board, since it is in review.

tasleson:
Looks like we are running into some stuff that Peter was warning us about.  For example, when we run the `loop_test_cache_size_change` test we fail here:

``` rust
cache.teardown(&dm).unwrap();
```
With:
```
Some(Os { code: 16, kind: Other, message: "Device or resource busy" })
```

If I put in a 2 second sleep before we do the teardown we complete without error.  Looks like when we set the IOCTL header flags to get the udev rules to run correctly, we are getting contention with the dm device and the udev infrastructure that is interrogating the device to create the symlinks etc.

Looks like we need to figure out how we want to deal with this...

mulkieran:
In a previous run on Travis, the assertion that failed was a few lines later:

```
---- cachedev::tests::loop_test_minimal_cache_dev stdout ----
	thread 'cachedev::tests::loop_test_minimal_cache_dev' panicked at 'assertion failed: `(left == right)`
  left: `91`,
 right: `0`', src/cachedev.rs:809:17
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
stack backtrace:
   0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
             at libstd/sys/unix/backtrace/tracing/gcc_s.rs:49
   1: std::sys_common::backtrace::_print
             at libstd/sys_common/backtrace.rs:71
   2: std::panicking::default_hook::{{closure}}
             at libstd/sys_common/backtrace.rs:59
             at libstd/panicking.rs:380
   3: std::panicking::default_hook
             at libstd/panicking.rs:390
   4: std::panicking::rust_panic_with_hook
             at libstd/panicking.rs:576
   5: std::panicking::begin_panic
             at libstd/panicking.rs:537
   6: std::panicking::begin_panic_fmt
             at libstd/panicking.rs:521
   7: devicemapper::cachedev::tests::test_minimal_cache_dev
             at src/cachedev.rs:809
   8: core::ops::function::Fn::call
             at /checkout/src/libcore/ops/function.rs:73
   9: devicemapper::loopbacked::test_with_spec
             at src/loopbacked.rs:129
  10: devicemapper::cachedev::tests::loop_test_minimal_cache_dev
             at src/cachedev.rs:850
  11: <F as alloc::boxed::FnBox<A>>::call_box
             at libtest/lib.rs:1462
             at /checkout/src/libcore/ops/function.rs:223
             at /checkout/src/liballoc/boxed.rs:788
  12: __rust_maybe_catch_panic
             at libpanic_unwind/lib.rs:102
```

Issue #256, opened 2018-01-29
New weird XFS debug output in our tests


mulkieran:
I don't recall seeing it before today.

mulkieran:
```
test thindev::tests::loop_test_snapshot_usage ... Clearing log and setting UUID
writing all SBs
new UUID = 328906fd-2482-49d5-a0b8-9a858ba4e518
```

mulkieran:
Did they mean to do that?

mulkieran:
Ok. It's just ```xfs_admin``` command-line call spitting out stuff in a snapshot usage test.

mulkieran:
I think this is closed by #335.

Issue #212, opened 2017-11-17
We should upgrade our test infrastructure to do cleanup


mulkieran:
We should do this in a similar way to stratisd's, but we might as well wait until stratisd's settles down a bit.

mulkieran:
But our devicemapper devices will not conform to a "stratis" naming scheme as they do with stratis tests. We should come up with a consistant naming scheme, and then implement it for all existing tests. I'm open to suggestions on the naming scheme, but I think it should not involve starting with stratis.

mulkieran:
#247 contains more info.

mulkieran:
Basic cleanup is mostly succeeding after tweak to Jenkins, so this is less urgent.

mulkieran:
I think this was closed by #309, et. al.

PR #299, opened 2018-04-23
Clean up between unit tests

tasleson:
We did something similar to this before for stratisd, adapted for devicemapper-rs.  This should address:

https://github.com/stratis-storage/devicemapper-rs/issues/212
~~https://github.com/stratis-storage/devicemapper-rs/issues/256~~

With this in place I get a lot less failures when running https://github.com/stratis-storage/devicemapper-rs/pull/296 locally (it reduces the number of failing error cases to examine).

UPDATED:
* Using @agrover change to use function instead of macro

### DIFF
--- a/src/loopbacked.rs
+++ b/src/loopbacked.rs
@@ -2,8 +2,8 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+use std::{io, panic};
 use std::fs::{File, OpenOptions};
-use std::io;
 use std::io::{Seek, SeekFrom, Write};
 use std::os::unix::prelude::AsRawFd;
 use std::path::{Path, PathBuf};
@@ -12,6 +12,7 @@ use loopdev::{LoopControl, LoopDevice};
 use tempdir::TempDir;
 
 use super::consts::{IEC, SECTOR_SIZE};
+use super::test_lib::clean_up;
 use super::types::{Bytes, Sectors};
 
 
@@ -119,12 +120,18 @@ fn get_devices(count: u8, dir: &TempDir) -> Vec<LoopTestDev> {
 /// Then, run the designated test.
 /// Then, take down the loop devices.
 pub fn test_with_spec<F>(count: u8, test: F) -> ()
-    where F: Fn(&[&Path]) -> ()
+    where F: Fn(&[&Path]) -> () + panic::RefUnwindSafe
 {
+    clean_up().unwrap();
+
     let tmpdir = TempDir::new("devicemapper").unwrap();
     let loop_devices: Vec<LoopTestDev> = get_devices(count, &tmpdir);
     let device_paths: Vec<PathBuf> = loop_devices.iter().map(|x| x.path()).collect();
     let device_paths: Vec<&Path> = device_paths.iter().map(|x| x.as_path()).collect();
 
-    test(&device_paths);
+    let result = panic::catch_unwind(|| test(&device_paths));
+    let tear_down = clean_up();
+
+    result.unwrap();
+    tear_down.unwrap();
 }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,7 @@ nix = "0.10"
 bitflags = "1"
 newtype_derive = "0.1"
 macro-attr = "0.2.0"
+mnt = "0.3.1"
 serde = "1"
 clippy = {version = "0", optional = true}
 error-chain = "0.11"
--- a/src/cachedev.rs
+++ b/src/cachedev.rs
@@ -715,6 +715,8 @@ use super::device::devnode_to_devno;
 use super::lineardev::LinearTargetParams;
 #[cfg(test)]
 use super::loopbacked::blkdev_size;
+#[cfg(test)]
+use super::test_lib::test_name;
 
 // Specified in kernel docs
 /// The minimum size recommended in the docs for a cache block.
@@ -731,7 +733,7 @@ pub fn minimal_cachedev(dm: &DM, paths: &[&Path]) -> CacheDev {
     assert!(paths.len() >= 2);
     let dev1 = Device::from(devnode_to_devno(paths[0]).unwrap().unwrap());
 
-    let meta_name = DmName::new("cache-meta").expect("valid format");
+    let meta_name = test_name("cache-meta").expect("valid format");
 
     // Minimum recommended metadata size for thinpool
     let meta_length = Sectors(4 * IEC::Ki);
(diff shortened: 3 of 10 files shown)


mulkieran:
This is blocked by #301.

mulkieran:
Restart clippy

src/shared_macros.rs
```diff
@@ -43,3 +43,17 @@ macro_rules! table {
         &$s.table
     }
 }
+
```
mulkieran: Like the stuff in util.rs, it would be better if this went in a separate file. Probably the same file as the stuff in util.rs.

src/shared_macros.rs
```diff
@@ -43,3 +43,17 @@ macro_rules! table {
...
+
+/// All things we create in the unit test will have this utilized so we know what to cleanup.
+/// If we change this we need to change the "tn" macro below too.
+#[cfg(test)]
+pub static DM_TEST_ID: &'static str = "_dm-rs_ut_delme";
```
mulkieran: I guess the "ut" stands for unit test? If so, maybe it should be dropped or possibly replace with "t", since it might extend its range to non-unit tests, somehow;.

src/shared_macros.rs
```diff
@@ -43,3 +43,17 @@ macro_rules! table {
...
+
+/// Generate a test name (tn) which can be identified later, so that is can be cleaned up.
```
mulkieran: why is it necessary for this to be a macro? AFAICT, there is no obstacle to making it a function.

tasleson: I initially tried utilizing a function and ran into borrow checker errors which would have required me to create a local instance for each usage so that it's lifetime was long enough.  If you can come up with a function that achieves the same change as this macro, please do share.

src/shared_macros.rs
tasleson: Sure

src/shared_macros.rs
tasleson: Yes, the function I created did indeed return a `String` and there is nothing wrong with that, but in doing so each and every instance would have required me to create a new local variable for the sole purpose of storing the name, so that it's lifetime existed for the duration of the function.  Yes, this isn't a big deal, but I thought creating a simple macro would be preferable as it minimizes the amount of code churn for the ~92 places the macro is being utilized.

What is your concern with leveraging a macro in this context?

tasleson: OK the concerns are:

* You cannot use a constant variable with `concat!` macro, requires string literal which causes us to have string in two different places which makes us susceptible to miss-match
* The table names are now required to be a string literal, which was true in existing code, but may not remain true

mulkieran: AFAIK:

The rational thing to do, if deciding to use a function to get this value, would be to make a ```DmNameBuf```, instead of a ```DmName```(as opposed to my ```AsRef``` suggestion, which would not help).

The reason the ```tn!``` macro works is because ```concat!``` and the macro are expanded at compile time, yielding a static str value. When ```DmName::new``` is applied to this static string, the lifetime of the resulting ```DmName``` is also static, thus lasting as long as the method. If the value is computed at runtime in any way, then it is necessary to either:

* use a let-statement that has scope has large as the ```DmName``` object requires, OR
* use a DmNameBuf

src/util.rs
```diff
@@ -15,3 +15,113 @@ pub fn align_to(num: usize, align_to: usize) -> usize {
 pub fn slice_to_null(slc: &[u8]) -> Option<&[u8]> {
     slc.iter().position(|c| *c == b'\0').map(|i| &slc[..i])
 }
+
+#[cfg(test)]
+use std::path::PathBuf;
+
+#[cfg(test)]
```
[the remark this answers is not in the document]
mulkieran: Looks like Class II, option (B) for singleton implementation might have been the best choice.

agrover:
I've explored the DmNameBuf approach, please take a look, and if it looks acceptable, squash. See #306.

src/util.rs
```diff
@@ -15,3 +15,113 @@ pub fn align_to(num: usize, align_to: usize) -> usize {
 pub fn slice_to_null(slc: &[u8]) -> Option<&[u8]> {
     slc.iter().position(|c| *c == b'\0').map(|i| &slc[..i])
 }
+
```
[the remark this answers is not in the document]
tasleson: Change incorporated, thanks

src/shared_macros.rs
tasleson: Change incorporated, thanks

src/shared_macros.rs
tasleson: Changed to `_dm-rs_test_delme`

src/shared_macros.rs
tasleson: No longer a macro :-)

tasleson:
@mulkieran I removed the stdout & stderr capture for executing external commands.  Will create a new PR with those changes as requested once this PR gets merged, thanks.

src/loopbacked.rs
```diff
@@ -119,12 +120,18 @@ fn get_devices(count: u8, dir: &TempDir) -> Vec<LoopTestDev> {
 /// Then, run the designated test.
 /// Then, take down the loop devices.
 pub fn test_with_spec<F>(count: u8, test: F) -> ()
-    where F: Fn(&[&Path]) -> ()
+    where F: Fn(&[&Path]) -> () + panic::RefUnwindSafe
 {
-    let tmpdir = TempDir::new("devicemapper").unwrap();
+    clean_up().unwrap();
+
+    let tmpdir = TempDir::new(test_name("devicemapper").unwrap().as_str()).unwrap();
```
mulkieran: I don't think any code cleans this up, so it doesn't make sense to have it generate its name with ```test_name```.

mulkieran: There is a larger issue. This is the unique place in the PR where the result of ```test_name``` is dug into, using the newly added ```as_str()``` in order to get the necessary str to pass to ```TempDir::new()```. If it is really necessary to name this temp dir using a test name, then it makes more sense to have ```test_name()``` just return ```String```, and pass that value to ```DmNameBuf``` constructor where a ```DmNameBuf``` is required, as appropriate.

agrover: maybe we should just come up with a different method for coming up with tempdir names, and not use `test_name`.

tasleson: I removed the `test_name ` use in `loopbacked.rs` as it's not needed.  However, we do need it in thindev.rs to ensure we umount the mounted test fs.  I added a new function `test_path` to be used in this context.  If this is unacceptable to folks, I will re-work so that we only have one function that returns a `String` and re-work all the usage to leverage that instead of returning `DmNameBuf`, `DmUuidBuf` etc.

agrover:
:+1: except I think we can drop the change to src/types.rs that added an `as_str` method, since we're doing something different for tempdir names now.

tasleson:
@agrover @mulkieran Mulhern has another PR with a number of other changes which I believe obsoletes this PR.  I think we can close this one and go forward with that one.

PR #307, opened 2018-05-02
RFC: Refactor API for future backwards compatibility

tasleson:
After trying to leverage the `event_nr` field for a different behavior change, it became apparent that each and every time the device mapper IOCTL interface is changed, it would potentially have far reaching changes to the `devicemapper-rs` API.  After reviewing the rust docs on API design, I tired to incorporate some changes which should make future backwards compatibility easier.  The main change is to introduce a `DmOptions` data type which conforms to a builder pattern to handle the ability to add to the interface over time.  I also considered making the `DM` struct fill this role, but it would be easy for a user to mistakenly call every method with the same options which typically isn't needed.  By passing the options as an argument to the function calls the user of the API has control on each function invocation.  The removal of the `DM` struct was more driven when I was debugging some other issues and found it incredibly unfortunate that I couldn't close the file during debug.  The cost of opening and closing the FD for each IOCTL is incurred with this change, but we could mitigate that with lazy-static initialization internally if needed in the future.  We could also keep the `DM` struct, but change it's use, so that it's mutable in `shared.rs` etc., so that it's future flexibility could be preserved.

Open question:
* The `shared.rs` public functions, should they expose the `DmOptions` too?  @mulkieran has been considering separating devicemapper-rs into two separate libraries, the `role` of the upper helper functions which are kind of specific to stratis could either hide this flexibility or bubble it up as a more generic library interface.

These are proposed changes, to spur some conversation on how we can create an API that will grow gracefully overtime.

### DIFF
--- a/src/cachedev.rs
+++ b/src/cachedev.rs
@@ -10,7 +10,6 @@ use std::str::FromStr;
 use super::device::Device;
 use super::deviceinfo::DeviceInfo;
 use super::dm;
-use super::dm_flags::DmFlags;
 use super::lineardev::{LinearDev, LinearDevTargetParams};
 use super::result::{DmError, DmResult, ErrorEnum};
 use super::shared::{DmDevice, TargetLine, TargetParams, TargetTable, device_create, device_exists,
@@ -421,7 +420,7 @@ impl DmDevice<CacheDevTargetTable> for CacheDev {
     }
 
     fn teardown(self) -> DmResult<()> {
-        dm::device_remove(&DevId::Name(self.name()), DmFlags::empty())?;
+        dm::device_remove(&DevId::Name(self.name()), None)?;
         self.cache_dev.teardown()?;
         self.origin_dev.teardown()?;
         self.meta_dev.teardown()?;
@@ -586,7 +585,7 @@ impl CacheDev {
     // Note: This method is not entirely complete. In particular, *_args values
     // may require more or better checking or processing.
     pub fn status(&self) -> DmResult<CacheDevStatus> {
-        let (_, status) = dm::table_status(&DevId::Name(self.name()), DmFlags::empty())?;
+        let (_, status) = dm::table_status(&DevId::Name(self.name()), None)?;
 
         assert_eq!(status.len(),
                    1,
--- a/src/cachedev.rs
+++ b/src/cachedev.rs
@@ -9,7 +9,7 @@ use std::str::FromStr;
 
 use super::device::Device;
 use super::deviceinfo::DeviceInfo;
-use super::dm::DM;
+use super::dm;
 use super::dm_flags::DmFlags;
 use super::lineardev::{LinearDev, LinearDevTargetParams};
 use super::result::{DmError, DmResult, ErrorEnum};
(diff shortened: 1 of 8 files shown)


mulkieran:
What particular Rust docs on API design?

tasleson:
> What particular Rust docs on API design?

@mulkieran https://rust-lang-nursery.github.io/api-guidelines/

tasleson:
I should probably have utilized a non-consuming builder (listed as preferred), but then you have to use a `let` statement to get long enough lifetime.  I tried it out both ways, ref. https://doc.rust-lang.org/1.0.0/style/ownership/builders.html

mulkieran:
Are there any particular parts of these Rust API recommendations that you recall being particularly illuminating in the context of this RFC?

agrover:
I like the second commit. DmFlags feels VERY low-level, using a builder pattern with DmOptions seems nicer. I wonder if the `Option`s within DmOptions are really needed -- why not just use 0? -- and if `Option` around `DmOption` is really needed too. Can't we just use DmOptions::new() or DmOptions::empty()?

The first commit, the comment assumes DM in shared.rs is immutable. Can we examine that? I'm not sure it should be -- we're making all sorts of changes to the system's DM configuration with a non-mut DM context! (When first writing dm-rs I just ignored this :smiley: ) There's the bigger question of going back and forth about if we care about opening/closing /dev/mapper frequently or not, but I wanted to raise the fact that immutable DM was perhaps an early mistake, as well, that is not set in stone.

mulkieran:
The first commit is approximately 1000 lines worth of change, but I can't find any explanation of what it does in the commit message. I'm waiting for some guidance as to what I should look for in the commit message before I begin review.

tasleson:
@agrover 

>  like the second commit. DmFlags feels VERY low-level, using a builder pattern with DmOptions seems nicer. I wonder if the Options within DmOptions are really needed -- why not just use 0? -- and if Option around DmOption is really needed too. Can't we just use DmOptions::new() or DmOptions::empty()?

Good suggestion, I think we could reduce some complexity by removing all the optional stuff.

> The first commit, the comment assumes DM in shared.rs is immutable. Can we examine that?

I tried adding a builder type interface to the dm structure.  The compiler lit up a ton of errors about mutability.  The immutable aspect is from the fact that the compiler is only looking for members in the structure to see if they are being changed, which they are not.  It's great that the rust compiler forces you to think about mutability and enforces it.  However, in some cases its not good in that it limits what you can do in the future in that if you're not using it in a mutable way now you can't do so in the future.  Wondering if there is a way to work around this...

agrover:
> Wondering if there is a way to work around this...

One can use a `Cell` or `RefCell` to modify an object through an immutable reference, so that's an option. But what I was also considering as a thought experiment was requiring `&mut DM` because even though the `File` doesn't change, one can make major changes to the kernel's DM configuration! So doesn't it logically make sense for DM needing to be `mut` if such changes to be possible?

agrover:
> The first commit is approximately 1000 lines worth of change, but I can't find any explanation of what it does in the commit message. I'm waiting for some guidance as to what I should look for in the commit message before I begin review.

It eliminates the need for `struct DM`, in favor of just opening /dev/mapper/control for each ioctl call. A pretty big change. (Everything except src/dm.rs changes are just callsite changes so they can probably be ignored for now.) Like I alluded to in my comment on Friday, it really depends on if it makes sense to require the caller to hold on to a context to use the library, or not. Is opening that file expensive? Well not really, but we just made a change to stratisd to minimize how many times we open it, so we obviously seem to care at some level. (But what we did there could've been viewed as more about eliminating having to pass struct DM everywhere.) If we decide we actually don't care, then this commit points out there's a significant simplification to dm-rs's external API that we can move towards.

agrover:
@tasleson both these commits are very significant. Any chance the second commit could get its own PR? This would let it possibly proceed and be merged while we discuss the merits of the first commit.

tasleson:
> @tasleson both these commits are very significant. Any chance the second commit could get its own PR? This would let it possibly proceed and be merged while we discuss the merits of the first commit.

@agrover  Yes, I can work on separating each of these commits into a separate PR.

tasleson:
Will reintroduce this change as two separate PRs.

PR #311, opened 2018-05-07
Remove DM struct

tasleson:
This change eliminates the need for struct DM, in favor of just opening
/dev/mapper/control for each ioctl call.  Previous stratisd changes have
introduced a singleton to allow code to fetch a reference as needed to
remove passing the dm structure reference around as much.  This change takes
it one last step by simply removing the structure all together and thus
removing the need for the singleton and passing the structure reference
around in the remaining places.  This change results in a significant
simplification to the public device mapper API design and reduces
public facing API coupling.

### DIFF
--- a/src/cachedev.rs
+++ b/src/cachedev.rs
@@ -9,7 +9,7 @@ use std::str::FromStr;
 
 use super::device::Device;
 use super::deviceinfo::DeviceInfo;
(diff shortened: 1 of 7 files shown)


tasleson:
This change seems a little too aggressive for folks, closing for now.

PR #312, opened 2018-05-07
Add DmOptions builder

tasleson:
Replace pub functions which took a DmFlags with a new DmOptions which utilizes a
builder pattern.  With this change we should be able to accommodate changes to
the device mapper structure over time, especially the IOCTL structure without
requiring new function calls to be created for every affected function.

This change incorporates suggestions from @agrover in this previous PR https://github.com/stratis-storage/devicemapper-rs/pull/307, mainly by removing the many uses of `Option`.

### DIFF
--- a/src/dm.rs
+++ b/src/dm.rs
@@ -17,6 +17,7 @@ use super::dm_ioctl as dmi;
 use super::dm_options::DmOptions;
 use super::errors::{Error, ErrorKind};
 use super::result::DmResult;
+
 use super::types::{DevId, DmName, DmNameBuf, DmUuid, Sectors, TargetTypeBuf};
 use super::util::{align_to, slice_to_null};
 
@@ -39,27 +40,11 @@ pub struct DM {
     file: File,
 }
 
-impl DM {
-    /// Create a new context for communicating with DM.
-    pub fn new() -> DmResult<DM> {
-        Ok(DM {
-            file: File::open(DM_CTL_PATH)
-                .map_err(|e| Error::with_chain(e, ErrorKind::ContextInitError))?,
-        })
-    }
-
-    /// Get the file within the DM context, likely for polling purposes.
-    pub fn file(&self) -> &File {
-        &self.file
-    }
-
-    fn generate_hdr(
-        id: Option<&DevId>,
-        options: &DmOptions,
-        allowable_flags: DmFlags,
-    ) -> dmi::Struct_dm_ioctl {
-        let clean_flags = allowable_flags & options.flags();
-        let event_nr = (options.cookie().bits() as u32) << 16;
+impl DmOptions {
+    /// Generate a header to be used for IOCTL.
+    fn hdr(&self, id: Option<&DevId>, allowable_flags: DmFlags) -> dmi::Struct_dm_ioctl {
+        let clean_flags = allowable_flags & self.flags();
+        let event_nr = (self.cookie().bits() as u32) << 16;
         let mut hdr: dmi::Struct_dm_ioctl = Default::default();
 
         hdr.version[0] = DM_VERSION_MAJOR;
@@ -73,12 +58,22 @@ impl DM {
 
         if let Some(id) = id {
             match *id {
-                DevId::Name(name) => Self::hdr_set_name(&mut hdr, name),
-                DevId::Uuid(uuid) => Self::hdr_set_uuid(&mut hdr, uuid),
+                DevId::Name(name) => DM::hdr_set_name(&mut hdr, name),
+                DevId::Uuid(uuid) => DM::hdr_set_uuid(&mut hdr, uuid),
             };
         };
 
         hdr
+    }
+}
+
+impl DM {
+    /// Create a new context for communicating with DM.
+    pub fn new() -> DmResult<DM> {
+        Ok(DM {
+            file: File::open(DM_CTL_PATH)
+                .map_err(|e| Error::with_chain(e, ErrorKind::ContextInitError))?,
+        })
     }
 
     fn hdr_set_name(hdr: &mut dmi::Struct_dm_ioctl, name: &DmName) -> () {
@@ -93,6 +88,11 @@ impl DM {
             unsafe { &mut *(&mut hdr.uuid as *mut [i8; DM_UUID_LEN] as *mut [u8; DM_UUID_LEN]) };
         let bytes = uuid.as_bytes();
         uuid_dest[..bytes.len()].clone_from_slice(bytes);
+    }
+
+    /// Get the file within the DM context, likely for polling purposes.
+    pub fn file(&self) -> &File {
+        &self.file
     }
 
     // Give this a filled-in header and optionally add'l stuff.
@@ -177,8 +177,7 @@ impl DM {
 
     /// Devicemapper version information: Major, Minor, and patchlevel versions.
     pub fn version(&self) -> DmResult<(u32, u32, u32)> {
-        let mut hdr: dmi::Struct_dm_ioctl =
-            Self::generate_hdr(None, &DmOptions::new(), DmFlags::empty());
+        let mut hdr = DmOptions::new().hdr(None, DmFlags::empty());
 
         self.do_ioctl(dmi::DM_VERSION_CMD as u8, &mut hdr, None)?;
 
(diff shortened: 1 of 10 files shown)


src/dm_options.rs
```diff
@@ -0,0 +1,50 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+use super::dm_flags::DmFlags;
+
+/// Encapsulates options for device mapper calls
+#[derive(Debug, Clone)]
```
agrover: suggest adding `Copy` to make life easier.

src/dm_options.rs
```diff
@@ -0,0 +1,50 @@
...
+        }
+    }
+
+    /// Synonym for new()
+    pub fn empty() -> DmOptions {
+        DmOptions::new()
+    }
+
+    /// Set the DmFlags value for option
+    pub fn set_flags(mut self, flags: DmFlags) -> DmOptions {
```
agrover: There's some ambiguity here...  if you call `set_flags` twice then only the DmFlags set last are set -- i.e. it's assigning to self.flags, not ORing bits.

It actually might be nice to OR here.

Or if not, at least make it clear in the docstring.

src/dm_options.rs
```diff
@@ -0,0 +1,50 @@
...
+        self.flags = flags;
+        self
+    }
+
+    /// Set the event_nr value for option
+    pub fn set_event_nr(mut self, event_nr: u32) -> DmOptions {
```
agrover: see above about whether this should actually be called cookie.

src/dm_options.rs
```diff
@@ -0,0 +1,50 @@
...
+        self.event_nr = event_nr;
+        self
+    }
+
+    /// Retrieve the flags value
+    pub fn get_flags(&self) -> DmFlags {
```
agrover: both for this and `get_event_nr`, leaving off the `get_` from getter method names is ok in Rust.

mulkieran:
I doubt that ```DmOptions``` could be said to use the "Builder pattern", as it only builds itself. It's just a bag of stuff with a ```new()``` function with 0 arguments and a bunch of setter and getter methods. I guess the idea here is to encapsulate all possible options that could be passed to a dm method in a single ```DmOptions``` object which could be extended as necessary with new attributes, in case the underlying library is extended.

src/dm_options.rs
mulkieran: I think not ORing is the part that makes ```DmOptions``` sort of be implementing the "Builder pattern", these functions are only supposed to be called, in sequence, when constructing a ```DmOptions``` object, so ORing doesn't really make sense. Probably ```DmOptions``` needs a docstring of its own, explaining how it should be used, and that would cover the problem.

src/dm_options.rs
mulkieran: If Copy is added, then that would be a decision to start passing ```DmOptions``` values by value everywhere. I think that would be fine.

src/dm.rs
```diff
@@ -62,6 +63,38 @@ impl DM {
         hdr.data_start = size_of::<dmi::Struct_dm_ioctl>() as u32;
     }
 
+    fn initialize_hdr_options(hdr: &mut dmi::Struct_dm_ioctl, flags: DmFlags, event_nr: u32) -> () {
```
mulkieran: ```initialize_hdr``` method, located above this method, can not be removed because some methods are still calling it. Is this an oversight, or is it necessary for some reason.

src/dm_options.rs
tasleson: @agrover @mulkieran Would we be limiting ourselves if we state that the type is copy-able?  Is it possible that future data members would prevent this from being true or is there always a way to mitigate that?

src/dm_options.rs
tasleson: I'll look at the docs, I think the upper 16 and lower 16 may need to be separated out in the interface.  Flags for the upper and an `u16` for the lower.

src/dm_options.rs
tasleson: :+1:

src/dm.rs
tasleson: Just an oversight, I think it can be removed.

src/dm_options.rs
```diff
@@ -0,0 +1,50 @@
...
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+use super::dm_flags::DmFlags;
+
+/// Encapsulates options for device mapper calls
+#[derive(Debug, Clone)]
+pub struct DmOptions {
+    flags: DmFlags,
+    event_nr: u32,
```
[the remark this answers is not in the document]
tasleson: I'll look @ docs, yes I think it would be good to break this out and leverage additional type(s) and flag values to reduce errors.

src/dm.rs
```diff
@@ -62,6 +63,38 @@ impl DM {
...
+
+    fn generate_hdr(id: Option<&DevId>,
+                    options: &DmOptions,
+                    allowable_flags: DmFlags)
+                    -> dmi::Struct_dm_ioctl {
+        let clean_flags = allowable_flags & options.get_flags();
+        let event_nr = options.get_event_nr();
+        let mut hdr: dmi::Struct_dm_ioctl = Default::default();
+
+        Self::initialize_hdr_options(&mut hdr, clean_flags, event_nr);
```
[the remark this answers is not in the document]
tasleson: Sure

src/dm_options.rs
mulkieran: All fields in the struct would have to be ```Copy``` for DmOptions to be ```Copy```. If that failing to be true is a concern, then ```DmOptions``` should not be ```Copy```.

tasleson:
Re-based

tasleson:
Removed the last commit which introduced `set_id`

src/dm_options.rs
```diff
@@ -0,0 +1,67 @@
...
+            flags: DmFlags::empty(),
+            cookie: DmCookie::empty(),
+        }
+    }
+
+    /// Set the DmFlags value for option.  Note this call is not additive in that it sets (replaces)
+    /// entire flag value in one call.  Thus if you want to incrementally add additional flags you
+    /// need to retrieve current and '|' with new.
+    ///
+    /// ```no_run
```
mulkieran: These doc tests seem to be run propertly, regardless, but in other doc tests three backticks are put at the bottom as well as the top of the doc test. Same point for doc test related to ```set_cookie```.

src/shared.rs
```diff
@@ -86,12 +90,13 @@ pub trait DmDevice<T: TargetTable> {
 
     /// Suspend I/O on the device. If flush is true, flush the device first.
     fn suspend(&mut self, dm: &DM, flush: bool) -> DmResult<()> {
-        let flags = if flush {
-            DmFlags::DM_SUSPEND
+        let mut options = DmOptions::new();
+        let options = if flush {
```
mulkieran: Nicer if:

```
let options = options.set_flags(if flush {DmFlags::DM_SUSPEND} else {DmFlags::DM_SUSPEND | DmFlags::DM_NOFLUSH});
```

or maybe

```
let options = DmOptions::new().set_flags(if flush {DmFlags::DM_SUSPEND} else {DmFlags::DM_SUSPEND | DmFlags::DM_NOFLUSH});
```

and get to dispense with the mut?

src/dm.rs
```diff
@@ -38,28 +40,40 @@ pub struct DM {
...
-            file: File::open(DM_CTL_PATH)
-                .map_err(|e| Error::with_chain(e, ErrorKind::ContextInitError))?,
-        })
-    }
-
-    /// Get the file within the DM context, likely for polling purposes.
-    pub fn file(&self) -> &File {
-        &self.file
-    }
+impl DmOptions {
```
mulkieran: Why is impl of DmOptions in two places?

src/dm.rs
```diff
@@ -38,28 +40,40 @@ pub struct DM {
...
         hdr.version[1] = DM_VERSION_MINOR;
         hdr.version[2] = DM_VERSION_PATCHLEVEL;
 
-        hdr.flags = flags.bits();
+        hdr.flags = clean_flags.bits();
+        hdr.event_nr = event_nr;
 
         hdr.data_start = size_of::<dmi::Struct_dm_ioctl>() as u32;
+
+        if let Some(id) = id {
```
mulkieran: It is a tiny bit more elegant to use:

```
if let Some(ref id) = id {
    match id { ...
```

tasleson: Indeed, thanks.

src/dm.rs
mulkieran: That probably is the best way, and I guess we might as well at this time. We can always expose it later if we have to.

tasleson:
Replacing this PR with https://github.com/stratis-storage/devicemapper-rs/pull/317 and squashed commits.

PR #314, opened 2018-05-16
Make clippy an allowed failure on Travis CI

mulkieran:
It has become too flaky to be allowed to gate a PR.


agrover:
:+1:

PR #323, opened 2018-05-31
Add DmOptions parameter to device_create

tasleson:
~~Current best approach for getting existing user space udev rules
to do what we need to get Stratis to be used for root fs and for
allowing Stratis backed XFS fs mounted in fstab.  This works by
setting a cookie value when we un-suspend the newly created thin
device which is utilized for newly created FS.~~

Current best approach for getting existing user space udev rules
to do what we need.  To allow Stratis to be used for root fs and for
allowing Stratis backed XFS fs mounted in fstab.  This works by
setting a cookie value when we resume the newly created thin
device which is utilized for newly created FS.

### DIFF
--- a/src/cachedev.rs
+++ b/src/cachedev.rs
@@ -481,7 +481,7 @@ impl CacheDev {
         }
 
         let table = CacheDev::gen_default_table(&meta, &cache, &origin, cache_block_size);
-        let dev_info = device_create(dm, name, uuid, &table)?;
+        let dev_info = device_create(dm, name, uuid, &table, &DmOptions::new())?;
 
         Ok(CacheDev {
             dev_info: Box::new(dev_info),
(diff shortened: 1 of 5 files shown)


agrover:
Code looks fine, except can we call it `device_create_udev_visible` or something? "process" is a word, both as a noun and verb, that are not accurate in this case I think.

tasleson:
> Code looks fine, except can we call it device_create_udev_visible or something? "process" is a word, both as a noun and verb, that are not accurate in this case I think.

@agrover Done!

mulkieran:
Why not use just ```device_create``` and pass it the correct ```DmOptions```?

tasleson:
Re-worked to add `DmOptions` parameter to `device_create`.

One of the reasons I didn't do something like this initially is that there are 3 different IOCTL calls in the implementation of `device_create`.  Each of those calls could specify a `DmOptions`, thus from an API perspective what makes the API most usable is not readily apparent and could be confusing to the caller or limiting over time.

agrover:
We all happy with this now?

PR #334, opened 2018-06-15
Set cookie for setup and snapshot

tasleson:
Previous PR:
#323

Introduced a cookie to get userspace to handle udev event, but
we left out the setup and snapshot code paths, so add those too.

### DIFF
--- a/src/thindev.rs
+++ b/src/thindev.rs
@@ -279,7 +279,13 @@ impl ThinDev {
             device_match(dm, &dev, uuid)?;
             dev
         } else {
-            let dev_info = device_create(dm, name, uuid, &table, &DmOptions::new())?;
+            let dev_info = device_create(
+                dm,
+                name,
+                uuid,
+                &table,
+                &DmOptions::new().set_cookie(DmCookie::DM_UDEV_PRIMARY_SOURCE_FLAG),
+            )?;
             ThinDev {
                 dev_info: Box::new(dev_info),
                 table,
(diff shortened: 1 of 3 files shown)


agrover:
The second commit looks fine, everybody likes more tests. For the first commit, well, it includes two things: setting the flag for snapshot/setup/new but also some refactoring. Clearly we need to do the first thing but the refactoring seems like a marginal win -- lines of code and and code paths go up.

Do you mind splitting the first commit into two, one that fixes the bug, and then one that does the refactoring?

tasleson:
The re-factor was done to reduce the number of code paths that are almost a direct copy & paste.  I very much subscribe to: https://en.wikipedia.org/wiki/Don%27t_repeat_yourself , but I will re-work to just fix the issue as you request.

tasleson:
I'll put the re-factoring into a different PR.

agrover:
Thanks. Yeah, just want to look at it by itself.

src/thindev.rs
```diff
@@ -432,6 +440,38 @@ mod tests {
...
+
+    // Return a hashmap of key-value pairs for udev entry.
+    fn get_udev_db_entry(dev_node_search: &PathBuf) -> Option<HashMap<String, String>> {
+        // Takes a libudev device entry and returns the properties as a HashMap.
+        fn device_as_map(device: &libudev::Device) -> HashMap<String, String> {
+            let rc: HashMap<_, _> = device
+                .properties()
+                .map(|i| {
+                    (
+                        String::from(i.name().to_str().expect("Unix is utf-8")),
```
mulkieran: Please just ```unwrap()``` here. That these property values can be decoded is far from a logical certainty. It seems like half the bugs that pyudev gets are about the fact that it's not even true. I'm not absolutely sure that the names can be decoded, although I feel that's likely.

tasleson: Done, thanks!

PR #335, opened 2018-06-18
Collect stdout & stderr from Command

tasleson:
Addresses https://github.com/stratis-storage/devicemapper-rs/issues/256

### DIFF
--- a/src/test_lib.rs
+++ b/src/test_lib.rs
@@ -3,15 +3,17 @@
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 use std::panic::catch_unwind;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
+use std::process::Command;
 use std::sync::{Once, ONCE_INIT};
 
 use mnt::get_submounts;
 use nix::mount::{MntFlags, umount2};
+use uuid::Uuid;
 
 use super::dm::DM;
 use super::dm_options::DmOptions;
-use super::result::DmResult;
+use super::result::{DmError, DmResult, ErrorEnum};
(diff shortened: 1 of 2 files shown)


agrover:
LGTM :+1:

src/test_lib.rs
```diff
@@ -38,6 +40,41 @@ pub fn test_string(name: &str) -> String {
     namestr
 }
 
+/// Execute command while collecting stdout & stderr.
+pub fn execute_cmd(cmd: &mut Command, error_msg: &str) -> DmResult<()> {
+    let result = cmd.output().unwrap();
```
mulkieran: Might as well just use "?" here. All these methods seem to be designed to return errors, except for ```get_dm```, which is anomalous and should be fixed so it behaves like the version in stratisd, which doesn't unwrap.

tasleson: Indeed, originally I didn't have the functions returning anything, but I switched them to return a result and I forgot to remove the unwrap, thanks!

src/test_lib.rs
```diff
@@ -38,6 +40,55 @@ pub fn test_string(name: &str) -> String {
     namestr
 }
 
+/// Execute command while collecting stdout & stderr.
+pub fn execute_cmd(cmd: &mut Command, error_msg: &str) -> DmResult<()> {
```
mulkieran: Try leaving out the ```error_msg``` parameter and just printing out the command that failed.

Sort of like in https://github.com/stratis-storage/stratisd/pull/988/commits/4050571d312b15e2ed04db5dacd0b5ad5b35d176.

This is only test code, but I think the principle still applies.

Issue #261, opened 2018-02-01
flakey target feature_args are not like feature args for other targets


mulkieran:
We handle them as if they were just a set of words like the other feature args. But they have more structure than that and must be treated accordingly. ```FIXME``` marks the spot.

PR #337, opened 2018-06-19
RFC: Reduce the amount of copy & pasted code

tasleson:
Some examples of what is duplicated and possible reductions.  Not sure how well received any of this will be.


mulkieran:
Issue #261 needs to be taken into account here.

Issue #202, opened 2017-11-07
Consider prettifying the output of some of our Debug impls


mulkieran:
Some of the structs that we want to implement Debug for include fields with type ```[T; N]``` where N is large enough that Debug isn't automatically implemented. It is easy enough to form these into something that implements Debug by making a reference to a Vec as: ```&thing.to_vec()```. But since N is big, the result is a long and ugly looking list, like: ```[1, 2, ..., 3, 4]``` except N elements long. This will be ugly to read, but at least gives the complete story, very important in debugging. Is there a way to make it less ugly, but just as informative?

Note: there is a pretty printing annotation for Debug: ```:#?```. This seems to reformat the text. But it doesn't really help, because it just turns a single very long row into a single very long column.

mulkieran:
We explicitly implement ```Debug``` in just two places: for range_macros and for the automatically generated dm_ioctl type. The problem identified in the issue: ```N``` too big for ```Debug``` to be implemented automatically has clearly gone away. Closing.

PR #339, opened 2018-07-09
Make Debug output for Struct_dm_ioctl more readable

agrover:
When formatting `Debug` output, print strings for Struct_dm_ioctl name and uuid fields, instead of arrays of u8, which are verbose and not that useful.

This brought to light the fact that having these arrays be `i8` is non-optimal, so fix that first.


agrover:
two follow-on commits to fix some clippy warnings and improve output of the dev field.

mulkieran:
This is related to, but does not fix: #202, i.e., we've just flipped from ugly hard-to-read output which tells the whole story to easy-to-read output which hides some information that we could end up wishing we had. But it may be more helpful at this time.

src/dm_ioctl.rs
```diff
@@ -132,9 +135,24 @@ impl fmt::Debug for Struct_dm_ioctl {
...
             .field("padding", &self.padding)
             .field("dev", &self.dev)
-            .field("name", &self.name.to_vec())
-            .field("uuid", &self.uuid.to_vec())
-            .field("data", &self.data.to_vec())
+            .field(
+                "name",
+                &slice_to_null(&self.name)
+                    .map(|s| String::from_utf8_lossy(s))
+                    .unwrap_or_else(|| Cow::Borrowed("")),
```
mulkieran: You can only have ```None``` here if there is no null-terminator to be found in the vec. Probably better to have some special way of indicating that, as
```<no value found>```, or something.

agrover: hmm. This should never happen, name and uuid are supposed to be null-terminated. Let's try `expect()`.

agrover:
* switched to using `expect()`
* Add a commit that omits unused/padding fields from Debug output

mulkieran:
If the only choice is an empty string as a defaykt or an ```expect()```, empty string is the right choice. Perhaps we know that dm always zeroes out all fields that it doesn't fill w/ an explicit value. And we also assume that dm code always puts a null-terminator on the end of the string that it writes in those fields. But, it doesn't make sense to crash just because some dm code has a bug, which is the effect of having an ```expect()``` instead of a default value. After all, we probably are using the Debug implementation because something went wrong. Hopefully it wasn't a badly written UUID on some unlikely code path in the kernel, but we can't be certain of that.

mulkieran:
Please change title to "Make Debug output for Struct_dm_ioctl more readable". "Improve" is a bit too broad, because in some debugging contexts, it will make the output useless. We just hope those debugging contexts are the less usual ones.

agrover:
Thanks! Merging.

PR #340, opened 2018-07-10
fixes for latest clippy

agrover:
These are good suggestions but they are also API changes.


agrover:
picked from #339

PR #346, opened 2018-07-11
Implement Debug for Sectors, DataBlocks, Bytes, and MetaBlocks

agrover:
The derived Debug implementation when pretty-printed results in e.g.:

Sectors (
0
)

This changes it to:

Sectors(0)

which isn't a huge change but is worth it because these types are used
very frequently.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -51,6 +51,18 @@ macro_rules! self_div {
     };
 }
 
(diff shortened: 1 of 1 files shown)


mulkieran:
devicemapper is not on Mergify, so I can do the honors.

PR #351, opened 2018-07-19
Remove some error-chain specific clippy allows

mulkieran:
We believe that the necessity has been removed by the recent error-chain
update: https://github.com/rust-lang-nursery/error-chain/issues/245


mulkieran:
Trivial so long as it passes tests, otherwise not.

PR #356, opened 2018-08-01
Add ThinPoolDev::set_low_water_mark() 

agrover:
This PR allows clients to set the low water mark for an existing ThinPoolDev, and also implements PartialEq so two otherwise identical ThinPoolDevs with different low water marks compare as equal. (Low water mark configures eventing behavior, rather than some value that is critical to the device working correctly.)

### DIFF
--- a/src/thinpooldev.rs
+++ b/src/thinpooldev.rs
@@ -26,13 +26,22 @@ use std::path::Path;
 
 const THINPOOL_TARGET_NAME: &str = "thin-pool";
 
-#[derive(Clone, Debug, Eq, PartialEq)]
+#[derive(Clone, Debug, Eq)]
 pub struct ThinPoolTargetParams {
     pub metadata_dev: Device,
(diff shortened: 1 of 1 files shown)


src/thinpooldev.rs
```diff
@@ -461,6 +470,35 @@ impl ThinPoolDev {
         )
     }
 
+    /// Set the low water mark to be something different than what was given
+    /// to ::new() or ::setup().
```
trgill: Just a suggestion.... maybe update the wording to indicate it isn't just what was given to new() or setup() - it may be called several times - updating from the previous value?

agrover:
@mulkieran thanks for taking a pass, makes sense and I've added that commit.

Shall I go ahead and follow our new procedure, and make a fresh, squashed PR? I'm assuming you approve of your own changes :smiley:

mulkieran:
@agrover Yes, that makes sense.

PR #357, opened 2018-08-01
Agrover add set lowat

mulkieran:
Changes are:
1. Use ```equivalent_tables``` instead of PartialEq override. This is a safer bet, for various reasons and is consistent w/ how it is done in cachedev.
2. Take advantage of the fact that tables can be cloned to be less verbose in new method.
3. Do not resume within the new method. Leave resuming to the caller. This is consistent w/ all the other methods and more expressive, which we need.


agrover:
contents merged in #358

PR #362, opened 2018-08-06
Test that ThinDev::destroy() removes the thin_id from use

agrover:

### DIFF
--- a/src/thindev.rs
+++ b/src/thindev.rs
@@ -882,7 +882,13 @@ mod tests {
         let thin_name = test_name("name").expect("is valid DM name");
 
         let td = ThinDev::new(&dm, &thin_name, None, tp.size(), &tp, thin_id).unwrap();
+        td.teardown(&dm).unwrap();
+
+        // This should work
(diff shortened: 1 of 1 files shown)


src/thindev.rs
```diff
@@ -869,6 +869,30 @@ mod tests {
...
+    fn test_thindev_destroy(paths: &[&Path]) -> () {
+        assert!(paths.len() >= 1);
+
+        let dm = DM::new().unwrap();
+        let tp = minimal_thinpool(&dm, paths[0]);
+
+        let thin_id = ThinDevId::new_u64(0).expect("is below limit");
+        let thin_name = test_name("name").expect("is valid DM name");
+
+        let td = ThinDev::new(&dm, &thin_name, None, tp.size(), &tp, thin_id).unwrap();
```
mulkieran: It would be more complete to have a ```setup()``` call right here as well. Otherwise, it's not definite that the subsequent failure of ```setup()``` is actually due to the ```destroy()``` call.

mulkieran:
Took it off the project since it's all done with and it was merged as #365.

PR #366, opened 2018-08-20
Add support for meta_low_water to ThinPoolWorkingStatus

agrover:
Upstream kernel has merged support for reporting the metadata low water
mark. The kernel sets this value, but userspace needs to know it in order
to tell when crossing the value has caused an event.

In some future API-breaking change we could just provide this value, but
for now it is available as an Option, to handle older kernels.


agrover:
Final PR is #368.

PR #371, opened 2018-09-04
Change teardown methods to take &mut self

agrover:
The problem with teardown taking self is that if it fails, we no longer
have ownership of the object. Instead, pass in a mutable reference. This
leaves the caller with the object if the teardown/destroy fails.


agrover:
Needed to fix https://github.com/stratis-storage/stratisd/issues/1137#issuecomment-417501033

agrover:
Sure, see the referenced stratisd PR, 1151.

PR #373, opened 2018-09-08
Add nix::Errno attribute to IoctlError

agrover:
In addition to having the DeviceInfo, it's also handy to know the errno
that caused the IoctlError, if available. Include this.

Change test locations since ErrorKind::IoctlError now has two fields.

Make errors public so external code has visibility into the Core error
types.

Signed-off-by: Andy Grover <agrover@redhat.com>

PR #374, opened 2018-09-13
Use primitive type-based chaining mechanism

mulkieran:
Due to restrictions in the standard library, using error-chain does not
allow most handy programmatic access to the elements of the chain.

Thus, it seems better to fall back on a simple type-based mechanism,
which is unfortunately more brittle but also more useful.

Make some description fields more descriptive.

Do not bother to remove error-chaining from test infrastructure, it is
not necessary to have programmatic access to it in the same way as to
a part of the actual library.

### DIFF
--- a/src/device.rs
+++ b/src/device.rs
@@ -10,7 +10,7 @@ use std::{fmt, io};
 use libc::{dev_t, major, makedev, minor};
 use nix::sys::stat::SFlag;
 
-use super::errors::{Error, ErrorKind};
+use super::errors::ErrorKind;
 use super::result::{DmError, DmResult};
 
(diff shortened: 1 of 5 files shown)


mulkieran:
This is an alternative to #373.

mulkieran:
Associated Stratis PR: https://github.com/stratis-storage/stratisd/pull/1179.

agrover:
ok I think this is nicer than #373, though very similar. Let's get this in and we can tag a new dm-rs release afterwards.

mulkieran:
This is in review-prelim because of the stray colons. But it looks like it's probably close to good to go, so I'll close it and open a final.

PR #375, opened 2018-09-13
Use primitive type-based chaining mechanism

mulkieran:
Due to restrictions in the standard library, using error-chain does not
allow most handy programmatic access to the elements of the chain.

Thus, it seems better to fall back on a simple type-based mechanism,
which is unfortunately more brittle but also more useful.

Make some description fields more descriptive.

Do not bother to remove error-chaining from test infrastructure, it is
not necessary to have programmatic access to it in the same way as to
a part of the actual library.


agrover:
I'm going to go ahead and expedite this one, there are things being blocked.

PR #378, opened 2018-09-27
Use clippy in the now recommended way

mulkieran:


mulkieran:
Might as well get this in before a new round of development starts.

PR #383, opened 2018-11-07
Keep track of ioctl buffer size

tasleson:
Addresses: https://github.com/stratis-storage/stratisd/issues/1308

### DIFF
--- a/src/dm.rs
+++ b/src/dm.rs
@@ -2,6 +2,7 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
+use std::cell::Cell;
 use std::fs::File;
 use std::mem::size_of;
 use std::os::unix::io::AsRawFd;
@@ -38,6 +39,7 @@ const MIN_BUF_SIZE: usize = 16 * 1024;
 /// Context needed for communicating with devicemapper.
 pub struct DM {
     file: File,
+    ioctl_buff_size: Cell<usize>, // Allow mutable field inside immutable struct
 }
 
 impl DmOptions {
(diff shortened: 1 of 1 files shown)


agrover:
The initial buffer size will suffice 99.9% of the time. For the remaining .1% the harm still seems very minor. It's hard to justify much code to fix such a small thing. How about something simpler, like bumping up the initial buffer size, and/or increasing the enlargement factor from 2x to 16x or something? A buffer that is allocated and then immediately freed is essentially "free" so we could just make it bigger, perhaps.

tasleson:
> The initial buffer size will suffice 99.9% of the time.

In larger configurations it's not sufficient and will result in multiple ioctl iterations each and every time we get a dm event.  This isn't a free operation as we found out that for large values of buffer size the re-size itself can take ~20 seconds, granted that was `(2**32 - 1)`.

>  It's hard to justify much code to fix such a small thing
``` diff
 src/dm.rs | 6 +++++-
 1 file changed, 5 insertions(+), 1 deletion(-)
```
This change is pretty small IMHO.

I guess what we could do is create a configuration which is our MAX supported configuration, find out how big the buffer needs to be, pad it a little bit and make that our default otherwise we will likely always end up in a situation where we are trying operations multiple times.

mulkieran:
> The initial buffer size will suffice 99.9% of the time. For the remaining .1% the harm still seems very minor. It's hard to justify much code to fix such a small thing.

This statement relies on assumptions about typical usage which we can't be confident of. Whenever we list the devices in a loop we use up about 64 bytes per device (rough and probably an over-approximation). We start out w/ 16 KiB for the buffer. That accommodates roughly 256 DM devices, which is not a lot. Please feel free to check my math.

So, typical users who like a lot of filesystems could end up always needing to reallocate when ```list_devices``` is invoked.

I don't think we can consider that a truly unusual problem. But, I also don't think we need to hurry up and find a solution _right now_.

Note that, as we do stress tests, ```list_devices``` is in the hot path. stratisd grabs the DM event, then it must find out the relevant device and hence it calls ```list_devices```. Then it hits check() for the relevant device, and then it resizes as necessary. We need to look at what we can do to make sure stratisd response is adequate, and it may be that making ```list_devices``` as fast as possible is part of the solution. But we shouldn't assume that yet.

Filing an issue and proceeding w/ caution seems like the correct approach to me.

PR #384, opened 2018-11-13
Add parse_value(), add description to parse_device(), and replace panic()s with Err()s

agrover:
Unify parsing of numbers into values with a function. Since we want to
say WHICH number parsing was bad, also take a description of what was being
parsed.

Use this to streamline places where parsing a number returns an error, but
also in places where parsing a number panics. My former position was that
a change in the kernel ABI (which is what dm status strings are) warranted
a panic but since it's actually easier to return a Result *and* it's
less verbose... why not play nice?

Change some unrelated spots to return an Err instead of panicing.

Add "desc" argument to parse_value() as well as parse_device(). This is
strictly to return error strings that give a little more info.

### DIFF
--- a/src/cachedev.rs
+++ b/src/cachedev.rs
@@ -83,9 +83,9 @@ impl FromStr for CacheTargetParams {
             return Err(DmError::Dm(ErrorEnum::Invalid, err_msg));
         }
 
-        let metadata_dev = parse_device(vals[1])?;
-        let cache_dev = parse_device(vals[2])?;
-        let origin_dev = parse_device(vals[3])?;
+        let metadata_dev = parse_device(vals[1], "metadata dev")?;
+        let cache_dev = parse_device(vals[2], "cache dev")?;
+        let origin_dev = parse_device(vals[3], "origin dev")?;
 
         let block_size = Sectors(parse_value(vals[4], "data block size")?);
         let num_feature_args: usize = parse_value(vals[5], "number of feature args")?;
--- a/src/cachedev.rs
+++ b/src/cachedev.rs
@@ -83,9 +83,9 @@ impl FromStr for CacheTargetParams {
             return Err(DmError::Dm(ErrorEnum::Invalid, err_msg));
         }
 
-        let metadata_dev = parse_device(vals[1], "metadata dev")?;
-        let cache_dev = parse_device(vals[2], "cache dev")?;
-        let origin_dev = parse_device(vals[3], "origin dev")?;
+        let metadata_dev = parse_device(vals[1], "metadata sub-device for cache target")?;
+        let cache_dev = parse_device(vals[2], "cache sub-device for cache target")?;
+        let origin_dev = parse_device(vals[3], "origin sub-device for cache target")?;
 
         let block_size = Sectors(parse_value(vals[4], "data block size")?);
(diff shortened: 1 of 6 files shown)


src/shared.rs
```diff
@@ -199,3 +199,16 @@ pub fn parse_device(val: &str) -> DmResult<Device> {
...
+
+/// Parse a value or return an error.
+pub fn parse_value<T>(val: &str, desc: &str) -> DmResult<T>
+where
+    T: FromStr,
+{
+    val.parse::<T>().map_err(|_| {
+        DmError::Dm(
+            ErrorEnum::Invalid,
+            format!("failed to parse value \"{}\" for: {}", val, desc),
```
mulkieran: I think my original choice of wording for error messages was easier to interpret than this one.

Example, my original wording:

```failed to parse value for number of feature args from "ag"```

new wording

```failed to parse value "ag" for: number of feature args```

I think it takes some work to understand the second.

File: src/cachedev.rs [hunk 2]
```diff
@@ -83,29 +83,12 @@ impl FromStr for CacheTargetParams {
...
-                    "failed to parse value for number of feature args from \"{}\"",
-                    vals[5]
-                ),
-            )
-        })?;
+        let metadata_dev = parse_device(vals[1], "metadata dev")?;
+        let cache_dev = parse_device(vals[2], "cache dev")?;
+        let origin_dev = parse_device(vals[3], "origin dev")?;
+
+        let block_size = Sectors(parse_value(vals[4], "block size")?);
```
mulkieran: "block size" -> "data block size"

src/device.rs
```diff
@@ -40,22 +41,9 @@ impl FromStr for Device {
...
-        })?;
-        let minor = vals[1].parse::<u32>().map_err(|_| {
-            DmError::Core(
-                ErrorKind::InvalidArgument(format!(
-                    "could not parse \"{}\" to obtain minor number",
-                    vals[0]
-                )).into(),
-            )
-        })?;
+        let major = parse_value(vals[0], "major")?;
```
mulkieran: These will return the wrong error type, so should be dropped.

mulkieran:
@agrover Please separate into two PRs, one for ```parse_device``` and one for ```parse_value```.

src/device.rs
agrover: ok

src/cachedev.rs
agrover: ok

agrover:
@mulkieran I've made suggested changes.

I broke out changes into 4 commits, which should make them easier to review independently, but think it best to stick to 1 PR since parse_device changes really aren't all that significant.

mulkieran:
I had a lot of difficulty looking at the ```parse_device``` changes mixed in with the others so I broke them out into #385.

agrover:
@mulkieran changes from #385 look fine -- if I pull those in here, can this PR move to final?

mulkieran:
> @mulkieran changes from #385 look fine -- if I pull those in here, can this PR move to final?

I still need to break out the ```parse_value``` changes and look at them separate from the ```parse_device``` changes.

agrover:
final PR is #387

PR #388, opened 2018-11-27
Fix ThinPoolDev::set_low_water_mark()

agrover:
Since we're passing new_table to table_load, that's the one we need
to set the low water mark for.

self.table gets set to new_table if the load succeeds.


mulkieran:
It looks to me like the explanation for the Jenkins failure is that the Jenkins machine previously ran a stratisd test, and didn't cleanup properly from that.

trgill:
After restart Jenkins CI passed.

PR #389, opened 2018-11-27
Get rid of stratis mentions in devicemapper

mulkieran:
They are almost certainly harmless, but they're not exactly correct either.


mulkieran:
This is having the same Jenkins problems as #388 . I think a reboot is in order.

PR #391, opened 2018-11-28
An experiment with using the failure crate in devicemapper

mulkieran:
This substitutes use of the failure crate for error-chain.

This uses the most complicated failure pattern, the combination of Error and ErrorKind. It breaks the recommendation of having ErrorKind constructors be nullary or only take simple arguments. As a consequence of this, comparison can not be done with equality, and the ```kind()``` method returns a reference to an ErrorKind, rather than a value.

~~It has a serious problem...if I make the Core Error implement std::Error it doesn't compile, complaining about an existing implementation of Fail. I don't know why this is.~~ This is due to the "blanket implementation" of ```Fail``` which I believe is a technical term. If the type implements certain traits and the crate imports ```failure``` then this blanket implementation will be applied. That implementation of ```Fail``` will be the prior one, and an explicit implementation of ```Fail``` will be the conflicting one.

~~One can legally drop the definition of ```fn description()``` from an error implementation in Rust 1.29.1, and I did so, but will that work for the earliest version of Rust that we are required to support? Maybe not.~~

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,6 +18,8 @@ newtype_derive = "0.1"
 macro-attr = "0.2.0"
 serde = "1"
 error-chain = "0.12"
+failure = "0.1.3"
+failure_derive = "0.1.3"
 
 [dev-dependencies]
 loopdev = "0.2"
--- a/src/device.rs
+++ b/src/device.rs
@@ -38,22 +38,22 @@ impl FromStr for Device {
         let vals = s.split(':').collect::<Vec<_>>();
         if vals.len() != 2 {
             let err_msg = format!("value \"{}\" split into wrong number of fields", s);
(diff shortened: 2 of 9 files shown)


mulkieran:
Maybe the docs are old, and I should try implementing Error but not Fail. Giving it a go.

mulkieran:
So far, so good.

mulkieran:
Now using clippy approved match syntax for constructors.

mulkieran:
Rerunning Jenkins, it may pass now that breaking backtrace version has been yanked.

mulkieran:
blocking while we consider whether to raise lowest supported version.

mulkieran:
We decided not to use failure and to make our own instead.

PR #392, opened 2018-11-28
Use matches crate for tests

mulkieran:


mulkieran:
This is very incomplete. It is obviously a good idea to use ```assert_matches!``` for its intended purpose in tests. But this would conflict w/ work on failure, so right now it's just partial.

mulkieran:
Superceded by #439.

PR #394, opened 2018-12-11
Fix unnecessary null return clippy errors

mulkieran:
It's easy to do.


mulkieran:
Now only 14 clippy lints, and they're all about new scoping!

PR #395, opened 2018-12-11
Run clippy in newest recommended way

mulkieran:
It seems we can use just clippy, according to clippy docs.


mulkieran:
We know it works. We can worry about it later.

mulkieran:
This is the now recommended way, it doesn't change our results from the previous way, might as well merge it.

PR #405, opened 2019-01-23
Use v2 of mergify configuration syntax

mulkieran:
v1 will become unsupported at the end of January.

### DIFF
--- a/.mergify.yml
+++ b/.mergify.yml
@@ -1,11 +1,12 @@
-rules:
-  default:
-    disabling_label: "do not merge"
-    protection:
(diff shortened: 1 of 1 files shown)


mulkieran:
It's trivial assuming that we trust the mergify developers to convert properly.

mulkieran:
Maybe we shouldn't have trusted their configuration transformation. I would have expected it to auto-merge. I will take another look.

mulkieran:
It doesn't auto merge via mergify because it is a change to the mergify configuration. merging by hand.

PR #407, opened 2019-01-25
Deny future-incompatible and unused lints

mulkieran:


mulkieran:
rebased. Hoping that Mergify condition now matches Travis task properly.

mulkieran:
This updates devicemapper-rs imports to be like stratisd. There is one difference between stratisd and devicemapper, though, which is that devicemapper exports a lot of names on its public interface. This situation raises a question which didn't exist in straitsd, but which I am asking for this PR.

The final commit is an example in which cachedev imports are made shorter by exporting public names directly from the public interface.

Is the approach of the final commit preferred, or should the PR be just left as is?

trgill:
I don't have a strong preference.

mulkieran:
> Looks fine to me.

Which way? Or are you saying that either way is fine w/ you?

mulkieran:
Ok. I've opened a final version #408 which drops the final commit.

PR #410, opened 2019-01-29
Tidy up *dev status methods more

### LINKED ISSUE #221: expect! in status() methods will probably lead to embarassment eventually
The argument for the expects is that devicemapper should be notified well ahead of time of any change to the format of the output. But, given that the docs in the kernel right now don't correctly or fully describe the current format, what we're really doing is iteratively discovering the format, and we should expect that process to be indefinite.

stratisd shouldn't crash because of unexpected values in status output, and it has already. Let's be more realistic and make status() methods return a Result type.

mulkieran:
The objects of this PR are:
* Better encapsulation.
* To be better prepared for a possible future refactoring that eliminates the concept of distinct kinds of devices entirely.
* To eliminate some needless boiler plate.
* To get rid of the last risky assertion involved in reading a single status line.
* It gets rid of some obsoleted comments.
* It parse one last thinpool status value, held metadata root.

This is a continuation of #387 and really resolves #221.

### DIFF
--- a/src/shared.rs
+++ b/src/shared.rs
@@ -229,8 +229,8 @@ pub fn get_status_line_fields<'a>(
         return Err(DmError::Dm(
             ErrorEnum::Invalid,
             format!(
-                "Insufficient number of fields for status; requires at least {}, found {}",
-                number_required, length
+                "Insufficient number of fields for status; requires at least {}, found only {} in status line \"{}\"",
+                number_required, length, status_line
             ),
         ));
     }
--- a/src/shared.rs
+++ b/src/shared.rs
@@ -216,3 +216,23 @@ where
         )
     })
 }
+
+/// Get fields for a single status line.
+/// Return an error if an insufficient number of fields are obtained.
+pub fn get_status_line_fields<'a>(
+    status_line: &'a str,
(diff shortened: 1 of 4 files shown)


mulkieran:
Should do a sanity check test for newly parsed thinpool status value.

mulkieran:
rebased.

mulkieran:
Uggg. I think I might have missed a spot.

src/shared.rs
```diff
@@ -216,3 +216,23 @@ where
...
+) -> DmResult<Vec<&'a str>> {
+    let status_vals = status_line.split(' ').collect::<Vec<_>>();
+    let length = status_vals.len();
+    if length < number_required {
+        return Err(DmError::Dm(
+            ErrorEnum::Invalid,
+            format!(
+                "Insufficient number of fields for status; requires at least {}, found {}",
+                number_required, length
+            ),
```
tasleson: Would it be valuable from a debugging standpoint to include the status_line value in the error message?

mulkieran: It would certainly be more complete.

mulkieran:
Superceded by #417.

PR #414, opened 2019-02-01
Build both source and tests on lowest supported compiler

mulkieran:
Compilation does not proceed very far at all on tests if using the build
target, but we do want to check tests.


mulkieran:
rebased.

PR #305, opened 2018-04-25
Strip out all non-core parts of devicemapper

mulkieran:
There are compelling aesthetic and practical reasons for splitting
existing library into two parts.

Signed-off-by: mulhern <amulhern@redhat.com>

PR #304, opened 2018-04-25
Disentangle outer devicemapper layer from core

mulkieran:
Signed-off-by: mulhern <amulhern@redhat.com>

PR #421, opened 2019-02-11
Make a core sub-module

mulkieran:
The core submodule should contain all the core parts of devicemapper. These generally are the parts that are involved with marshalling and un-marshalling the ioctl results. The outer layer is the type-heavy layer that distinguishes between different types of devices.

This PR accomplishes two things:
* It puts all the core parts of devicemapper into a submodule. Importantly for further work, only methods in the core submodule can raise core errors, and they only raise those core errors.
* It makes the extraction of status values in the core, i.e., within ```dm::parse_table``` yield rawer, more primitive types. So, the first value in a table line is not converted to a Sector, it's left as a number, and so forth. This allows a better separation of the submodule from the outer module, and will allow the library better access to the raw values, if that proves to be useful for error reporting.

It also allows closing #304 and #305 , which tried to separate this library into two separate libraries. At this time, a separate module gives most of the advantages of two separate libraries and is less risky and hopefully lower maintenance.


mulkieran:
rebased.

mulkieran:
rebased.

mulkieran:
rebased.

mulkieran:
Fixed up formatting.

mulkieran:
PR of stratisd against this version of devicemapper: https://github.com/stratis-storage/stratisd/pull/1455.

Issue #434, opened 2019-03-05
devicemapper should dereference non-null pointers directly instead of using as_ref


mulkieran:
There are idioms like:

```
(result.as_ptr() as *const dmi::Struct_dm_name_list)
                        .as_ref()
                        .expect("pointer to own structure result can not be NULL")
```
and others that are much less clear about why the expect() should hold.

These can be simplified to:

```&*(result.as_ptr() as *const dmi::Struct_dm_name_list)```

because it is devicemapper that is managing that memory, and so the pointer is certainly not null.

theyoyojo:
So, changes like this?
<pre>

diff --git a/src/core/dm.rs b/src/core/dm.rs
index 94e68b9..32972fc 100644
--- a/src/core/dm.rs
+++ b/src/core/dm.rs
@@ -218,9 +218,7 @@ impl DM {
 
             loop {
                 let device = unsafe {
-                    (result.as_ptr() as *const dmi::Struct_dm_name_list)
-                        .as_ref()
-                        .expect("pointer to own structure result can not be NULL")
+                    &*(result.as_ptr() as *const dmi::Struct_dm_name_list)
                 };
 
                 let slc = slice_to_null(&result[size_of::<dmi::Struct_dm_name_list>()..])

mulkieran:
Yes. There are actually two parts to this problem.

1. The first part is the transformation, exactly like that.
2. The second part is actually verifying that it is true that the pointer can not be NULL. We always have an expect("<some string>") there, so the code is always asserting that the pointer can not be NULL. Sometimes we have the completely wrong reason, e.g., "assume all parsing succeeds", unfortunately.
But the second half of the problem is making sure that our assumption that the pointer can never be NULL is correct. And that involves finding out where that slice actually originated. Oh, and sometimes

* ```as_ptr()``` -> ```as_mut_ptr()```
* ```as_ref()``` -> ```as_mut()```

also.

mulkieran:
@theyoyojo Better hold off on this. I'm not so sure about non-NULLness anymore.

theyoyojo:
nice

Issue #433, opened 2019-03-05
devicemapper undecided about whether values from kernel should be valid utf8 or not


mulkieran:
Either

* all should be considered possibly invalid, and converting bytes from utf8 should possibly return an error, OR
* all should be considered certainly valid, and should use an expect()

I prefer the first option very strongly.

Sometimes the code takes the second option, but often the code takes the middle road, and uses ```from_utf8_lossy``` which makes no sense, as substituting special characters for unrepresentable characters in the result of an ioctl makes no sense. Probably this was considered safest, since it will not fail, so there is no need to accomodate error returns, but it it feels less risky than an assertion.

mulkieran:
Note that converting from utf8 is almost invariably preceded by a call to ```slice_to_null()```. This is always expected to succeed, which is also not really safe.

mulkieran:
We could start out by forcing ```slice_to_null``` to return an Error if it can't find a null character.

mulkieran:
Closed by #633. All strings from the kernel are now handled consistently when being decoded, and the possibility of an error is accounted for.

PR #430, opened 2019-02-26
Abstract getting unique status result into a single method

mulkieran:
Related: https://github.com/stratis-storage/stratisd/issues/794.
Related: https://github.com/stratis-storage/devicemapper-rs/issues/221.

The goals of this PR are:
* Remove some assertions, and replace them with errors, since the assertions are _not_ logically necessary.
* Make reading the target type when parsing the ioctl a bit safer in order to
* Extend the implementation of FromStr for the various status types by having it parse the target type. This bring FromStr for status types in line with FromStr for param types.
* Get rid of some unnecessary code duplication.

The changes in behavior are:
* Where previously an error in parsing might result in an assertion failure, now the error would result in an error return with a more helpful message.
* If it was ever possible for a target type to have a trailing space at the end, now it is not. This makes checking equality of target types a less chancy proposition.
* If a status line for an unexpected target type is parsed, a failure will occur when the target type is compared with the expected one, rather than later, when the actual fields of the status are parsed with incorrect expectations.

EDIT: Comments:
* https://github.com/stratis-storage/devicemapper-rs/pull/430#issuecomment-468005022
* https://github.com/stratis-storage/devicemapper-rs/pull/430#issuecomment-468009341
* https://github.com/stratis-storage/devicemapper-rs/pull/430#issuecomment-468014275
* https://github.com/stratis-storage/devicemapper-rs/pull/430#issuecomment-468024204
properly belong to issue #431, rather than to this PR.

PR #435, opened 2019-03-05
Trim the end of the target type string

mulkieran:
To be on the safe side. Also, it looks tidier because it matches the code
block following.


mulkieran:
This is cherry-picked from #430 , because it is really more or less orthogonal. I think there are a bunch of other things that should be done to improve this method, but these things apply to a lot of other methods, so I filed a few issues instead: #434 , #433 .

PR #439, opened 2019-03-08
Modify assert!(match {}) calls in test cases to use assert_matches!()…

theyoyojo:
… from crate::assert_matches.

Signed-off-by: Joel Savitz <jsavitz@redhat.com>

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,7 @@ loopdev = "0.2"
 tempfile = "3.0.1"
 libudev = "0.2.0"
 libmount = "0.1.11"
+matches = "0.1.8"
 
 [dev-dependencies.uuid]
 version = "0.7"
--- a/src/cachedev.rs
+++ b/src/cachedev.rs
@@ -428,7 +428,7 @@ impl FromStr for CacheDevStatus {
                     rest_start_index + 1,
                     val,
                     "needs check",
-                ))
+                ));
             }
         };
 
--- a/src/core/dm.rs
+++ b/src/core/dm.rs
@@ -794,10 +794,12 @@ mod tests {
             .unwrap();
 
         let new_uuid = test_uuid("example-9999999999").expect("is valid DM uuid");
-        assert!(match dm.device_rename(&name, &DevId::Uuid(&new_uuid)) {
-            Err(DmError::Core(Error(ErrorKind::IoctlError(_, _), _))) => true,
-            _ => false,
-        });
+
+        assert_matches!(
+            dm.device_rename(&name, &DevId::Uuid(&new_uuid)),
+            Err(DmError::Core(Error(ErrorKind::IoctlError(_, _), _)))
+        );
+
         dm.device_remove(&DevId::Name(&name), &DmOptions::new())
             .unwrap();
     }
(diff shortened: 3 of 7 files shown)


theyoyojo:
@mulkieran How does this look?

src/cachedev.rs
```diff
@@ -428,7 +428,7 @@ impl FromStr for CacheDevStatus {
                     rest_start_index + 1,
                     val,
                     "needs check",
-                ))
+                ));
```
mulkieran: Note: This edit is mandatory to satisfy rustfmt. This is inexplicable, why wasn't it mandatory last time?

theyoyojo: I'm not sure Do you want me to make the change in a new PR or update this one?

theyoyojo: I'm not sure. Do you want me to make the change in a new PR or update this one?

Cargo.toml
```diff
@@ -18,6 +18,7 @@ newtype_derive = "0.1"
 macro-attr = "0.2.0"
 serde = "1"
 error-chain = "0.12"
+assert_matches = "1.3.0"
```
mulkieran: Whoops! We have to use the matches crate, and it has to be a dev-dependency. See equivalent stuff in stratisd PR I referred you to.

src/cachedev.rs
mulkieran: You can't make the changes or rustfmt will complain. I can not figure out why. This is for everybody else's benefit, because it looks like a stray edit.

Cargo.toml
mulkieran: You can make this change in this PR and just force push, if you prefer.

src/cachedev.rs
mulkieran: I think it's probably because rustfmt is not idempotent. yapf (a Python formatter), has that problem as well.

theyoyojo:
OK. Anything else need tweaking?

src/cachedev.rs
theyoyojo: Does this affect this particular PR or is this a more general concern?

Cargo.toml
theyoyojo: Done.

mulkieran: The Cargo.toml file is where you specify the crate. See equivalent in stratisd PR: https://github.com/stratis-storage/stratisd/pull/1344/files#diff-80398c5faae3c069e4e6aa2ed11b28c0R37. For example, you'll want to make it a dev-dependency, since it is only used for tests.

src/cachedev.rs
mulkieran: It's a complete mystery! And we've never seen it before, we'll just keep an eye on it.

Cargo.toml
theyoyojo: Ah, my bad, I changed that but I must have forgotten to squash it into this force commit. Will fix shortly...

theyoyojo: This is resolved.

theyoyojo:
Crate fix:
assert_matches (1.3.0/dependency) -> matches (0.1.8/dev-dependency)

theyoyojo:
All tests function identically (11 failures, etc) after this single insertion/deletion

mulkieran:
Ok! I'm putting this in final review.

theyoyojo:
:+1:

PR #440, opened 2019-03-18
Change recommended development compiler to 1.32.0

mulkieran:
Just wondering what's in store...


mulkieran:
Looks like it'll be no problem at all.

PR #441, opened 2019-03-26
Use edition 2018

mulkieran:
Set the edition to 2018.

Run "cargo fix" on the source code and accept the changes.
Remove extern declarations that the compiler complains about.
Also, remove some extern declarations that the compiler does not complain about, but which
can be removed anyway.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,6 +9,7 @@ repository = "https://github.com/stratis-storage/devicemapper-rs"
 readme = "README.md"
 keywords = ["Linux", "device", "mapper", "libdm", "storage"]
 license = "MPL-2.0"
+edition = "2018"
 
 [dependencies]
 libc = "0.2.36"
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-DENY = "-D warnings -D future-incompatible -D unused -D ellipsis-inclusive-range-patterns"
+DENY = "-D warnings -D future-incompatible -D unused -D bare-trait-objects -D ellipsis-inclusive-range-patterns"
 
 ${HOME}/.cargo/bin/cargo-expand:
 	cargo install cargo-expand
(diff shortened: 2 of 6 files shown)


mulkieran:
rebased.

mulkieran:
This leaves only three unshiftable externs:

```
extern crate macro_attr;
extern crate newtype_derive;
extern crate error_chain;
```

but we are aiming to get rid of our dependencies on all three anyway.

mulkieran:
Note that #445 will get rid of ```macro_attr``` and ```newtype_derive```, leaving only ```error_chain```.

mulkieran:
Back to in progress, I don't expect any further comment for a while.

mulkieran:
I'm dubious about removing macro extern declarations. If we do, we must supply the names of macros that are not called explicitly in our source, but must be called transitively by the source of the crate that we import. I don't know that they have any obligation to keep those names stable, so this would result in a pesky versioning problem.

mulkieran:
rebased.

cuviper:
> I'm dubious about removing macro extern declarations. If we do, we must supply the names of macros that are not called explicitly in our source, but must be called transitively by the source of the crate that we import.

If those transitive macro calls are not prefixed with proper `$crate::` or such -- which they *can't* do if they're supporting older rustc -- then I would just keep your `#[macro_use] extern crate foo;` for those.

> I don't know that they have any obligation to keep those names stable, so this would result in a pesky versioning problem.

It's a gray area, especially if they have `#[doc(hidden)]` those internal pieces, then there's a social expectation that they can feel free to change those, never mind semver. This doesn't play well with explicit macro imports though.

mulkieran:
No longer using explict macro imports, keeping the extern declarations.

mulkieran:
rebased.

mulkieran:
Superceded by #466.

PR #446, opened 2019-04-02
Show expanded macros on Travis

mulkieran:
This is a convenience rather than a check, although it may be moved up to a check at some time. It is the case that cargo-expand takes longer than any other of the Travis tasks, however, the configuration file is set up so that the cumulative result is reported as soon as the required tasks are done, so this should not slow work at all.


mulkieran:
Blocked by #447 .

mulkieran:
rebased and updated.

PR #447, opened 2019-04-03
Use #[cfg(test)] for range_macros tests

mulkieran:
This was missing from commit f4f2fdd53940d94431c5022798360b4ea7.

### DIFF
--- a/src/range_macros.rs
+++ b/src/range_macros.rs
@@ -236,6 +236,7 @@ macro_rules! checked_add {
     };
 }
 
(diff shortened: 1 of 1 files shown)


mulkieran:
Note that the absence of the compiler annotation did not mean that the tests were not run. It's absence meant that the test module was compiled into the build build and probably the release build as well (unless optimizations would somehow remove it). With the annotation, the devicemapper library is 1% smaller.

PR #450, opened 2019-04-09
Removed the macro_attr and newtype_derive dependencies from definition of units

waltdisgrace:
-Wrote macros to implement the traits Add, Sub, AddAssign, SubAssign, Deref, From
-Replaced calls to Newtype macros with these new macros
-Removed the macro_attr and newtype_derive dependencies from definition of units
-Build size reduced by 2.36% (was 9241466, is now 9023322)

### DIFF
--- a/src/range_macros.rs
+++ b/src/range_macros.rs
@@ -5,29 +5,24 @@
 // An omnibus macro that includes all simple macros.
 macro_rules! range {
     ($T: ident, $display_name: expr) => {
-        macro_attr! {
-            #[derive(Clone, Copy, Default, Eq, Ord, PartialEq, PartialOrd)]
-            /// A type for $T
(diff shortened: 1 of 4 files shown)


src/range_macros.rs
```diff
@@ -42,6 +37,69 @@ macro_rules! self_div {
...
+            }
+        }
+    };
+}
+
+macro_rules! deref {
+    ($T:ident) => {
+        impl Deref for $T {
+            type Target = u64;
+            fn deref(&self) -> &Self::Target {
```
mulkieran: One change: ```&Self::Target``` -> ```u64```.

mulkieran:
@bgurney-rh , @drckeefe , @trgill: please take a look.

PR #457, opened 2019-04-22
Remove dependency on error_chain in source

mulkieran:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,7 @@ nix = "0.13"
 bitflags = "1"
 serde = "1"
 error-chain = "0.12"
+backtrace = "0"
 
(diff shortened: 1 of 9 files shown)


mulkieran:
This will not pass tests due to a conflict between error_chain and the new attempt at tests.

mulkieran:
This is expected to pass all tests, because the newly defined error type is Send. It can only be Send if the error's that it picks up are also send.

mulkieran:
This change reduces the size of the debug build by about 5%, but I'm not sure why.

mulkieran:
This was being used as test bed for experiments in error management; right now we just want to get error_chain out. Superseded by #671 .

PR #461, opened 2019-04-24
Upgrade recommended development compiler to 1.33

mulkieran:
* Upgrade development compiler to 1.33
* Group Travis tasks into distinct sets with related features and comment
* Add a docs test. Such a test already exists for stratisd, might as well have one here.


mulkieran:
@trgill This PR has a "default" task, which I don't remember from before.

PR #467, opened 2019-05-10
Look at a file that has expanded code w/ a ? operator

mulkieran:
This is just a quick way to get a look at the results of expandng the ? operator, or else to prove that it's a built in and isn't expanded. Please ignore.


mulkieran:
Sigh. ? is not expanded.

PR #471, opened 2019-05-16
Improve doc comments generated by macros

waltdisgrace:
I removed the doc comments containing $T.

I began to add my own doc comments, but I do not know how to make the
comments change depending on the $display_name.

How should I go about doing this?

### DIFF
--- a/src/range_macros.rs
+++ b/src/range_macros.rs
@@ -2,11 +2,12 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-// An omnibus macro that includes all simple macros.
+/// An omnibus macro that includes all simple macros.
 macro_rules! range {
     ($T: ident, $display_name: expr) => {
         #[derive(Clone, Copy, Default, Eq, Ord, PartialEq, PartialOrd)]
-        /// A type for $T
+
+        /// A struct upon which operations can be performed.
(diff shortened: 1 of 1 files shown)


mulkieran:
What do you see when you run '> make docs'?

mulkieran:
You'll have to intuit how and where the rust doc command picks up ```///``` comments in order to generate docs. Play w/ it slowly and carefully, making one change, and then inspecting the generated docs to see what the effect is.

waltdisgrace:
@mulkerian See this related issue: https://github.com/rust-lang/rust/issues/37903

waltdisgrace:
@mulkieran See this related issue: rust-lang/rust#37903

waltdisgrace:
@mulkieran See related issue: https://github.com/rust-lang/rust/issues/37903

mulkieran:
> @mulkieran See related issue: [rust-lang/rust#37903](https://github.com/rust-lang/rust/issues/37903)

Whoops! Thanks for finding this out.

So we have to back up and set our sites a bit lower.

1. Remove those nonsense comments w/ the $Ts that show up in the docs, because they do just look like garbage.
2. I'll try to rewrite that explanation of the type I sent you so that it's completely generic, and you can figure out how to include that.

But I don't think there's much more that can be done that would make a positive difference, so we might as well stop there.

mulkieran:
Obsoleted, since we want to make a bigger change in the implementation.

PR #474, opened 2019-05-22
Try out xenial

mulkieran:
This is expected to fail; according to the docs xenial does not supply Rust.


mulkieran:
Well! It turns out that xenial does support Rust. The failure was in a single test among the sudo tests.

PR #475, opened 2019-06-04
Update nix to 0.14

cuviper:
This is needed to fix breaking changes in libc on arm and s390x.
See: https://github.com/nix-rust/nix/pull/1055

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,7 @@ edition = "2018"
 
 [dependencies]
 libc = "0.2.36"
-nix = "0.13"
+nix = "0.14"
 bitflags = "1"
 serde = "1"
 error-chain = "0.12"


mulkieran:
ok to test

mulkieran:
Right now, in order for stratisd to compile on Rawhide, there needs to be a new version of devicemapper-rs w/ this new requirement, AFAICT. Other notes:  nix 0.14 requires libc 0.2.55. But to really fix things, stratisd should also require nix 0.14. The reason there is no outstanding PR for that change as well, is that it can be fixed in rawhide by just updating the spec file.

cuviper:
See https://github.com/stratis-storage/stratisd/pull/1552

mulkieran:
> See [stratis-storage/stratisd#1552](https://github.com/stratis-storage/stratisd/pull/1552)

Egad! Yes, I see it now.

PR #483, opened 2019-07-15
Implemented flakey target optional feature parameters as an enum 

waltdisgrace:
Resolves: #261.

### DIFF
--- a/src/lineardev.rs
+++ b/src/lineardev.rs
@@ -227,7 +227,6 @@ impl FromStr for FlakeyTargetParams {
             for x in 0..num_feature_args {
                 if vals[x] == "drop_writes" {
                     result.push(DropWrites);
-                    print!("DEBUG: result: {:?}", result);
                 } else if vals[x] == "error_writes" {
                     result.push(ErrorWrites);
                 } else if vals[x] == "corrupt_bio_byte" {
@@ -818,7 +817,7 @@ mod tests {
             .collect::<HashSet<_>>();
         assert_eq!(result.feature_args, expected);
     }
-    
+
     #[test]
     fn loop_test_duplicate_segments() {
         test_with_spec(1, test_duplicate_segments);
--- a/src/lineardev.rs
+++ b/src/lineardev.rs
@@ -210,16 +210,17 @@ impl FromStr for FlakeyTargetParams {
     type Err = DmError;
 
     fn from_str(s: &str) -> DmResult<FlakeyTargetParams> {
-        fn parse_direction(vals: &str) -> Direction {
+        fn parse_direction(val: &str) -> DmResult<Direction> {
             let result: Direction;
-            if vals == "r" {
+            if val == "r" {
                 result = Direction::Reads;
-            } else if vals == "w" {
+            } else if val == "w" {
                 result = Direction::Writes;
             } else {
-                panic!();
+                let err_msg = format!("Expected r or w, got {}", val);
+                return Err(DmError::Dm(ErrorEnum::Invalid, err_msg));
             }
-            result
+            Ok(result)
         }
 
         fn parse_feature_args(vals: &[&str], num_feature_args: usize) -> Vec<FeatureArg> {
@@ -231,7 +232,7 @@ impl FromStr for FlakeyTargetParams {
                     result.push(ErrorWrites);
                 } else if vals[x] == "corrupt_bio_byte" {
                     let byte: u64 = vals[x + 1].parse::<u64>().unwrap();
-                    let direction: Direction = parse_direction(vals[x + 2]);
+                    let direction: Direction = parse_direction(vals[x + 2]).unwrap();
                     let value: u8 = vals[x + 3].parse::<u8>().unwrap();
                     let flags: u64 = vals[x + 4].parse::<u64>().unwrap();
                     result.push(CorruptBioByte(byte, direction, value, flags));
--- a/src/lineardev.rs
+++ b/src/lineardev.rs
@@ -819,6 +819,48 @@ mod tests {
     }
 
     #[test]
+    fn test_flakey_target_params_corrupt_bio_byte_and_drop_writes() {
+        let result = "flakey 8:32 0 16 2 6 corrupt_bio_byte 32 r 1 0 drop_writes"
+            .parse::<FlakeyTargetParams>()
+            .unwrap();
+        let expected = [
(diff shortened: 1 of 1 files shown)


src/lineardev.rs
```diff
@@ -81,6 +81,51 @@ impl TargetParams for LinearTargetParams {
...
+    /// Read I/O is handled correctly.
+    ErrorWrites,
+    /// corrupt_bio_byte <Nth_byte> <direction> <value> <flags>:
+    ///
+    ///	During <down interval>, replace <Nth_byte> of the data of
+    ///	each matching bio with <value>.
+    ///
+    /// <Nth_byte>: The offset of the byte to replace.
+    ///             Counting starts at 1, to replace the first byte.
+    /// <direction>: Either 'r' to corrupt reads or 'w' to corrupt writes.
```
mulkieran: It probably makes sense to define a simple Direction enum, with a constructor for 'r' and 'w' rather than just use a ```char``` type here. Please go ahead and add that and other necessary associated changes as a separate commit.

src/lineardev.rs
```diff
@@ -229,10 +229,10 @@ impl FromStr for FlakeyTargetParams {
                 } else if vals[x] == "error_writes" {
                     result.push(ErrorWrites);
                 } else if vals[x] == "corrupt_bio_byte" {
-                    let byte: u64 = vals[x + 1].parse::<u64>().unwrap();
-                    let direction: Direction = parse_direction(vals[x + 2]).unwrap();
-                    let value: u8 = vals[x + 3].parse::<u8>().unwrap();
-                    let flags: u64 = vals[x + 4].parse::<u64>().unwrap();
+                    let byte: u64 = parse_value(vals[x + 1], "byte")?;
+                    let direction: Direction = parse_direction(vals[x + 2])?;
+                    let value: u8 = parse_value(vals[x + 4], "value")?;
```
mulkieran: index wrong.

src/lineardev.rs
```diff
@@ -140,33 +199,48 @@ impl fmt::Display for FlakeyTargetParams {
...
+
+        fn parse_feature_args(vals: &[&str], num_feature_args: usize) -> DmResult<Vec<FeatureArg>> {
+            let mut result: Vec<FeatureArg> = Vec::new();
+            for x in 0..num_feature_args - 1 {
+                if vals[x] == "drop_writes" {
+                    result.push(DropWrites);
+                } else if vals[x] == "error_writes" {
+                    result.push(ErrorWrites);
+                } else if vals[x] == "corrupt_bio_byte" {
+                    let byte: u64 = parse_value(vals[x + 1], "byte")?;
```
mulkieran: It's not easy to be that useful w/ the description, but "offset" would definitely be better than "byte", "byte offset" even better.

src/lineardev.rs
```diff
@@ -140,33 +199,90 @@ impl fmt::Display for FlakeyTargetParams {
...
     }
 }
 
+use crate::lineardev::FeatureArg::{CorruptBioByte, DropWrites, ErrorWrites};
+
 impl FromStr for FlakeyTargetParams {
     type Err = DmError;
 
     fn from_str(s: &str) -> DmResult<FlakeyTargetParams> {
+        fn parse_direction(val: &str, desc: &str) -> DmResult<Direction> {
```
mulkieran: You should implement the whole functionality of ```parse_direction``` by having Direction implement ```FromStr```. You'll find a bunch of examples of ```FromStr``` implementation.

src/lineardev.rs
```diff
@@ -140,33 +199,90 @@ impl fmt::Display for FlakeyTargetParams {
...
+                            })
+                            .and_then(|s| parse_value::<u64>(*s, "byte offset"))?;
+
+                        let direction = vals_iter
+                            .next()
+                            .ok_or({
+                                let err_msg = "corrupt_bio_byte takes 4 parameters";
+                                DmError::Dm(ErrorEnum::Invalid, err_msg.to_string())
+                            })
+                            .and_then(|s| parse_direction(*s, "direction"))?;
```
mulkieran: If you implement ```FromStr``` for ```Direction``` then this line will become ```parse_value::<Direction>``` rather than ```parse_direction```.

src/lineardev.rs
```diff
@@ -81,6 +81,79 @@ impl TargetParams for LinearTargetParams {
...
+
+impl FromStr for Direction {
+    type Err = DmError;
+    fn from_str(val: &str) -> DmResult<Direction> {
+        if val == "r" {
+            Ok(Direction::Reads)
+        } else if val == "w" {
+            Ok(Direction::Writes)
+        } else {
+            let err_msg = format!("Failed to parse from input \"{}\", expected r or w", val);
```
mulkieran: You'll need to make this error message make sense again, I think it got hacked up a bit for a quick compile.

src/lineardev.rs
```diff
@@ -140,33 +213,69 @@ impl fmt::Display for FlakeyTargetParams {
...
-    ///		 'w' is incompatible with drop_writes.
-    ///    <value>: The value (from 0-255) to write.
-    ///    <flags>: Perform the replacement only if bio->bi_opf has all the
-    ///	     selected flags set.
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         write!(f, "{} {}", FLAKEY_TARGET_NAME, self.param_str())
     }
 }
 
+use crate::lineardev::FeatureArg::{CorruptBioByte, DropWrites, ErrorWrites};
```
mulkieran: Please drop this line entirely, and identify the individual tags with their full name, as: ``FeatureArg::CorruptBioByte```.

src/lineardev.rs
```diff
@@ -81,6 +81,79 @@ impl TargetParams for LinearTargetParams {
...
+                f,
+                "corrupt_bio_byte {} {} {} {}",
+                offset, direction, value, flags
+            ),
+        }
+    }
+}
+
 /// Target params for flakey target
 // FIXME: Refine feature args handling. Reading the docs indicates that flakey
```
mulkieran: Whoops, for got this! You can get rid of the FIXME comment, because you have fixed the problem.

jbaublitz: We'll take a look at this tomorrow.

jbaublitz: Resolved in #492.

Issue #484, opened 2019-07-16
If the number of feature arguments is 0, then it is not mandatory


waltdisgrace:
We must modify our code, because we are currently treating the number of feature arguments as mandatory.

Evidence for the number of feature arguments not being mandatory:

1. The [DM source code](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/drivers/md) implements feature parsing such that supplying no feature arguments is allowed.
2. The [DM documentation](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/device-mapper) does not indicate that the number of feature arguments is mandatory.
3. Testing out various dmsetup commands shows that even when no number of feature arguments is supplied, a device can be created. See example below:

[gchin@to-be-determined devicemapper-rs]$ lsblk
NAME   MAJ:MIN RM   SIZE RO TYPE MOUNTPOINT
sda      8:0    0 119.2G  0 disk 
├─sda1   8:1    0     1G  0 part /boot
├─sda2   8:2    0  11.9G  0 part [SWAP]
├─sda3   8:3    0    50G  0 part /
├─sda4   8:4    0     1K  0 part 
└─sda5   8:5    0  56.3G  0 part /home
sdb      8:16   0 111.8G  0 disk 
├─sdb1   8:17   0     4G  0 part 
├─sdb2   8:18   0     5G  0 part 
└─sdb3   8:19   0     5G  0 part 
[gchin@to-be-determined devicemapper-rs]$ sudo dmsetup create flakey1 --table '0 8388608 flakey 8:17 0 10 10'
[gchin@to-be-determined devicemapper-rs]$ lsblk
NAME        MAJ:MIN RM   SIZE RO TYPE MOUNTPOINT
sda           8:0    0 119.2G  0 disk 
├─sda1        8:1    0     1G  0 part /boot
├─sda2        8:2    0  11.9G  0 part [SWAP]
├─sda3        8:3    0    50G  0 part /
├─sda4        8:4    0     1K  0 part 
└─sda5        8:5    0  56.3G  0 part /home
sdb           8:16   0 111.8G  0 disk 
├─sdb1        8:17   0     4G  0 part 
│ └─flakey1 253:0    0     4G  0 dm   
├─sdb2        8:18   0     5G  0 part 
└─sdb3        8:19   0     5G  0 part

mulkieran:
[...]

PR #488, opened 2019-07-23
Properly handle lack of feature arguments

waltdisgrace:
Resolves: #484 .

### DIFF
--- a/src/lineardev.rs
+++ b/src/lineardev.rs
@@ -167,7 +167,7 @@ impl FromStr for FlakeyTargetParams {
     type Err = DmError;
 
     fn from_str(s: &str) -> DmResult<FlakeyTargetParams> {
-        let mut vals = s.split(' ').collect::<Vec<_>>();
+        let vals = s.split(' ').collect::<Vec<_>>();
 
         if vals.len() < 5 {
             let err_msg = format!(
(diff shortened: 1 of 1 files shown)


src/lineardev.rs
```diff
@@ -192,6 +192,10 @@ impl FromStr for FlakeyTargetParams {
         let up_interval = parse_value(vals[3], "up interval")?;
         let down_interval = parse_value(vals[4], "down interval")?;
 
+        if vals.len() == 5 {
+            vals.push("0");
+        }
+
         let num_feature_args: usize = parse_value(vals[5], "number of feature args")?;
```
mulkieran: Your original approach, something like:

```let num_feature_args = if vals.len() == 5 { 0 } else { parse_value(vals[5], ...) };```

is really better.

One thing that makes it better is that you don't have to make ```vals``` mutable. This reduced the burden on the reader by quite a bit; if ```vals``` is mutable then they have to inspect the entire body of the method for places where the code changes it in some way.

File: src/lineardev.rs [hunk 2]
```diff
@@ -709,6 +711,22 @@ mod tests {
         ld.teardown(&dm).unwrap();
     }
 
+    #[test]
+    fn test_flakey_target_params_zero() {
+        let result = "flakey 8:32 0 16 2 0"
+            .parse::<FlakeyTargetParams>()
+            .unwrap();
+        let expected = [].iter().cloned().collect::<HashSet<_>>();
```
mulkieran: This is a bit of a long way around to get an empty HashSet. Just use ```HashSet::new``` as,

```assert_eq!(result.feature_args, HashSet::new());```

Same for test below.

mulkieran:
Superceded by #489.

PR #491, opened 2019-07-25
Implemented flakey target optional feature parameters as an enum

### LINKED ISSUE #261: flakey target feature_args are not like feature args for other targets

waltdisgrace:
Resolves #261.


waltdisgrace:
Needs to be redone, closing.

PR #493, opened 2019-07-29
Properly handle lack of thinpool feature args

waltdisgrace:
Addresses the thinpool part of #484.

### DIFF
--- a/src/thinpooldev.rs
+++ b/src/thinpooldev.rs
@@ -63,9 +63,9 @@ impl FromStr for ThinPoolTargetParams {
     fn from_str(s: &str) -> DmResult<ThinPoolTargetParams> {
         let vals = s.split(' ').collect::<Vec<_>>();
 
-        if vals.len() < 6 {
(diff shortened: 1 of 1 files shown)


mulkieran:
Please do a rebase off master...you don't need a new PR for that.

src/thinpooldev.rs
```diff
@@ -63,7 +63,7 @@ impl FromStr for ThinPoolTargetParams {
     fn from_str(s: &str) -> DmResult<ThinPoolTargetParams> {
         let vals = s.split(' ').collect::<Vec<_>>();
 
-        if vals.len() < 6 {
+        if vals.len() < 5 {
             let err_msg = format!(
                 "expected at least 6 values in params string \"{}\", found {}",
```
bgurney-rh: The check above changed from 6 to 5; should the message change to "Expected at least 5 values in params string..."?

waltdisgrace: Yes, thank you.

PR #497, opened 2019-08-02
Require error-chain 0.12.1

mulkieran:
Related: https://github.com/stratis-storage/project/issues/57.


mulkieran:
rebased.

mulkieran:
Wierd, probably unrelated test failure. It probably makes sense to investigate the test failure, though, as it seems to have recurred.

mulkieran:
Looks like the device is just not fully resumed when we try to destroy it.

PR #504, opened 2019-10-01
Add an audit task to the Makefile and to Travis CI

mulkieran:


mulkieran:
This checks a subset of what the stratisd task already checks, since stratisd depends on devicemapper-rs, but since it is an independent library, we might as well audit it.

mulkieran:
rebased.

PR #505, opened 2019-10-02
Make blkgetsize64 private

mulkieran:


mulkieran:
Failed! Will just document as in stratisd.

mulkieran:
This one doesn't call for review since exactly the same as in stratisd. I'll merge it as soon as it passes.

Issue #463, opened 2019-04-25
Improve docs generated by macros


mulkieran:
The doc comments document the macros, not the code that the macros generate. So, all types generated by the range! macro explain their type very tersely, ```A type for $T```.

See https://docs.rs/devicemapper/0.27.0/devicemapper/struct.Sectors.html.

The macro should generate text that documents the Sectors types, when it's the Sectors type being built, etc.

mulkieran:
Note that what we want here is that the macro should generate good docs for the four types that it constructs: Sectors, Bytes, MetaBlocks, DataBlocks. In any case, $T should not appear in the document comments.

mulkieran:
This is dropping down a bit in importance. Should we implement these w/out using macros at all, just as parameterized types? It might be almost exactly equivalent in terms of behavior and amount of code generated.

mulkieran:
Related: #101.

PR #512, opened 2019-10-25
Fix autogenerated docs

jbaublitz:
Resolves: #463.

### DIFF
--- a/src/range_macros.rs
+++ b/src/range_macros.rs
@@ -4,9 +4,11 @@
 
 // An omnibus macro that includes all simple macros.
 macro_rules! range {
-    ($T: ident, $display_name: expr) => {
+    ( $(#[$comment:meta])? $T: ident, $display_name: expr) => {
+        $(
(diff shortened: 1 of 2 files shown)


mulkieran:
Please just get rid of the comment on ```checked_add``` which has a related problem while you're here, so you can make a clean sweep of it.

jbaublitz:
Instead of having to use `#[allow(missing_docs)]`, I modified the docs slightly to be less specific but still somewhat helpful. Does that work?

mulkieran:
> Instead of having to use `#[allow(missing_docs)]`, I modified the docs slightly to be less specific but still somewhat helpful. Does that work?

Aaah, that explains why it was documented in the first place! Yes, that's best.

jbaublitz:
Should I rebase and squash?

jbaublitz:
@mulkieran Good to go.

jbaublitz:
Should I merge?

PR #520, opened 2019-11-25
Do not make test_lib public outside the testing module

mulkieran:
This PR tries to improve encapsulation by making test_lib private in the testing submodule and removing one publicity modifier that becomes not-useful with the preceding change.

### DIFF
--- a/src/testing/mod.rs
+++ b/src/testing/mod.rs
@@ -5,7 +5,7 @@
 //! Modules that support testing.
 
 mod loopbacked;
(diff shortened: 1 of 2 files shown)


mulkieran:
rebased.

mulkieran:
> Looks like the Travis test hit a spurious failure on the first attempt (EBUSY while trying to run loop_test_basic); retrying results in a successful test.

Thanks!

PR #516, opened 2019-11-11
Add list_test_devices method

waltdisgrace:
This is a preliminary PR that will address #465.

Issue #465, opened 2019-04-26
Robustify our tests so they don't assume only the devices they created exist


mulkieran:
The problem is that if there are other devicemapper devices, not created by the tests, hanging around, these tests will fail. Presumably they were always run previously in a context where there were no other devicemapper devices about.

In some cases, all this would need would be to adapt the condition tested to not require that assumption, but there might be some more difficult ones.

Two examples which go wrong are:
* core::dm::tests::sudo_test_list_devices
* core::dm::tests::sudo_test_list_devices_empty

mulkieran:
Need to talk about how to test this first.

waltdisgrace:
I can think of several ways to approach this:

1. Before performing any tests, find out which, if any, devices already exist. Store identifiers for these devices in a data structure or file. Perform all tests in such a way that all of the devices corresponding to the stored identifiers are **not** taken into account.
2. Before performing any tests, find out which, if any, devices already exist. Store all necessary information for re-creating these devices in a data structure or file. Remove all of the devices. Perform all of the tests. Re-create the devices that you removed.
3. Whenever you perform a test that creates a device, store an identifier for said device in a data structure or file. Perform all tests that don't create devices in such a way that **only** the devices corresponding to the stored identifiers are taken into account.
4. Rely on a standardized test device nomenclature for distinguishing which devices are relevant for testing.

waltdisgrace:
Closed by: #521.

PR #521, opened 2019-11-26
Add list_test_devices method

waltdisgrace:
Supersedes #516.
Addresses #465.

### DIFF
--- a/src/testing/test_lib.rs
+++ b/src/testing/test_lib.rs
@@ -17,13 +17,25 @@ use nix::mount::{umount2, MntFlags};
 use uuid::Uuid;
 
 use crate::{
-    core::{DevId, DmNameBuf, DmOptions, DmUuidBuf, DM},
+    core::{DevId, Device, DmNameBuf, DmOptions, DmUuidBuf, DM},
     result::{DmError, DmResult, ErrorEnum},
     units::Bytes,
 };
 
 static INIT: Once = Once::new();
 static mut DM_CONTEXT: Option<DM> = None;
+
+impl DM {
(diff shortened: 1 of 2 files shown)


src/testing/test_lib.rs
```diff
@@ -17,14 +17,25 @@ use nix::mount::{umount2, MntFlags};
...
+    core::{DevId, Device, DmNameBuf, DmOptions, DmUuidBuf, DM},
     result::{DmError, DmResult, ErrorEnum},
     units::Bytes,
 };
 
 static INIT: Once = Once::new();
 static mut DM_CONTEXT: Option<DM> = None;
 
+impl DM {
+    /// Returns a list of tuples containing DM test device names, a Device, which
```
mulkieran: I guess I missed this...this header comment is a word-for-word copy of the header comment on the ```DM::list_devices()``` method. That doesn't make sense, because this method does something different from what the ```DM::list_devices()``` method does (otherwise there would be no justification for its existence). Rather than copying and pasting explain what this method does _in terms_ of what ```DM::list_devices()``` does. Start with something like:

```Returns a subset of the devices returned by list_devices``` and continue from there, explaining what the modification is.

waltdisgrace: The change here is "containing DM test device names" vs. "containing DM device names."

mulkieran: > The change here is "containing DM test device names" vs. "containing DM device names."

Aaah, I missed that, many others will. Make the comment this method's very own; the best way to do it is to compose the comment from scratch, keeping in mind _exactly_ what the method does and why it exists.

PR #523, opened 2019-12-17
Use fully qualified name for all macro non-prelude types

bgurney-rh:
Related: #522

### DIFF
--- a/src/range_macros.rs
+++ b/src/range_macros.rs
@@ -30,7 +30,7 @@ macro_rules! range {
 
 macro_rules! self_div {
     ($T:ident) => {
-        impl Div<$T> for $T {
+        impl std::ops::Div<$T> for $T {
             type Output = u64;
             fn div(self, rhs: $T) -> u64 {
                 self.0 / *rhs
@@ -41,7 +41,7 @@ macro_rules! self_div {
 
 macro_rules! add {
     ($T:ident) => {
-        impl Add<$T> for $T {
+        impl std::ops::Add<$T> for $T {
             type Output = $T;
             fn add(self, rhs: $T) -> $T {
                 $T(self.0 + *rhs)
@@ -52,7 +52,7 @@ macro_rules! add {
 
 macro_rules! sub {
     ($T:ident) => {
-        impl Sub<$T> for $T {
+        impl std::ops::Sub<$T> for $T {
             type Output = $T;
             fn sub(self, rhs: $T) -> $T {
                 $T(self.0 - *rhs)
@@ -63,7 +63,7 @@ macro_rules! sub {
 
 macro_rules! add_assign {
     ($T:ident) => {
-        impl AddAssign<$T> for $T {
+        impl std::ops::AddAssign<$T> for $T {
             fn add_assign(&mut self, rhs: $T) {
                 *self = $T(self.0 + *rhs)
             }
@@ -73,7 +73,7 @@ macro_rules! add_assign {
 
 macro_rules! sub_assign {
     ($T:ident) => {
-        impl SubAssign<$T> for $T {
+        impl std::ops::SubAssign<$T> for $T {
             fn sub_assign(&mut self, rhs: $T) {
                 *self = $T(self.0 - *rhs)
             }
@@ -83,7 +83,7 @@ macro_rules! sub_assign {
 
 macro_rules! deref {
     ($T:ident) => {
-        impl Deref for $T {
+        impl std::ops::Deref for $T {
             type Target = u64;
             fn deref(&self) -> &u64 {
                 &self.0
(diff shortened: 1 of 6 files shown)


src/units.rs
```diff
@@ -2,11 +2,7 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
-use std::{
-    fmt,
-    iter::Sum,
-    ops::{Add, AddAssign, Deref, Div, Mul, Rem, Sub, SubAssign},
-};
+use std::{fmt, iter::Sum};
```
mulkieran: If you give Sum and Display the treatment in the range_marcros.rs macros all these std imports ought to go away (I think).

bgurney-rh: Ah, that's interesting in how it's defined, so I see "impl fmt::Display" in the macro definitions.

I also see "impl fmt::Debug"; should that be qualified as well?

mulkieran: Yes, it really should. I don't know why I'm surprised that Debug is not in the prelude, but evidently it isn't.

bgurney-rh:
I added the qualifiers for id_macros.rs and shared_macros.rs.

src/range_macros.rs
```diff
@@ -299,12 +299,7 @@ macro_rules! checked_add {
 #[cfg(test)]
 mod tests {
 
-    use std::{
-        fmt,
-        iter::Sum,
-        ops::{Add, AddAssign, Deref, Div, Mul, Rem, Sub, SubAssign},
-        u64,
-    };
+    use std::{fmt, u64};
```
mulkieran: I bet fmt use will go away when you make suggested changes.

bgurney-rh: Done; should be in the next force-push (after this most recent one; I forgot to run "cargo fmt").

bgurney-rh:
That last series was a bit dizzying, so here's a diff of it just in case:
```
diff --git a/src/core/types.rs b/src/core/types.rs
index 725c1e0..8780999 100644
--- a/src/core/types.rs
+++ b/src/core/types.rs
@@ -9,7 +9,7 @@ use crate::{
         deviceinfo::{DM_NAME_LEN, DM_UUID_LEN},
         errors::ErrorKind,
     },
-    result::{DmError, DmResult},
+    result::DmError,
 };
 
 /// An error function to construct an error when creating a new string id.
diff --git a/src/id_macros.rs b/src/id_macros.rs
index 6741237..58ca076 100644
--- a/src/id_macros.rs
+++ b/src/id_macros.rs
@@ -35,13 +35,13 @@ macro_rules! str_check {
 macro_rules! str_id {
     ($B:ident, $O:ident, $MAX:ident, $err_func:ident) => {
         /// The borrowed version of the DM identifier.
-        #[derive(Debug, PartialEq, Eq, Hash)]
+        #[derive(std::fmt::Debug, PartialEq, Eq, std::hash::Hash)]
         pub struct $B {
             inner: str,
         }
 
         /// The owned version of the DM identifier.
-        #[derive(Debug, Clone, PartialEq, Eq, Hash)]
+        #[derive(std::fmt::Debug, Clone, PartialEq, Eq, std::hash::Hash)]
         pub struct $O {
             inner: String,
         }
@@ -49,7 +49,7 @@ macro_rules! str_id {
         impl $B {
             /// Create a new borrowed identifier from a `&str`.
             #[allow(clippy::new_ret_no_self)]
-            pub fn new(value: &str) -> DmResult<&$B> {
+            pub fn new(value: &str) -> $crate::result::DmResult<&$B> {
                 if let Some(err_msg) = str_check!(value, $MAX - 1) {
                     return Err($err_func(&err_msg));
                 }
@@ -72,7 +72,7 @@ macro_rules! str_id {
         }
 
         impl std::fmt::Display for $B {
-            fn fmt(&self, f: &mut fmt::Formatter) -> std::fmt::Result {
+            fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
                 write!(f, "{}", &self.inner)
             }
         }
@@ -80,7 +80,7 @@ macro_rules! str_id {
         impl $O {
             /// Construct a new owned identifier.
             #[allow(clippy::new_ret_no_self)]
-            pub fn new(value: String) -> DmResult<$O> {
+            pub fn new(value: String) -> $crate::result::DmResult<$O> {
                 if let Some(err_msg) = str_check!(&value, $MAX - 1) {
                     return Err($err_func(&err_msg));
                 }
@@ -111,11 +111,11 @@ macro_rules! str_id {
 
 #[cfg(test)]
 mod tests {
-    use std::{fmt, iter, ops::Deref};
+    use std::{iter, ops::Deref};
 
     use crate::{
         core::errors::{Error, ErrorKind},
-        result::{DmError, DmResult},
+        result::DmError,
     };
 
     fn err_func(err_msg: &str) -> DmError {
```

bgurney-rh:
Okay, I think this is (hopefully) ready.

src/range_macros.rs
bgurney-rh: It did.

PR #524, opened 2019-12-19
Use fully qualified name for all macro non-prelude types

bgurney-rh:
Resolves: #522

Supercedes: #523

### DIFF
--- a/src/id_macros.rs
+++ b/src/id_macros.rs
@@ -49,7 +49,7 @@ macro_rules! str_id {
         impl $B {
             /// Create a new borrowed identifier from a `&str`.
             #[allow(clippy::new_ret_no_self)]
-            pub fn new(value: &str) -> DmResult<&$B> {
+            pub fn new(value: &str) -> $crate::result::DmResult<&$B> {
                 if let Some(err_msg) = str_check!(value, $MAX - 1) {
                     return Err($err_func(&err_msg));
                 }
@@ -71,8 +71,8 @@ macro_rules! str_id {
             }
         }
 
-        impl fmt::Display for $B {
-            fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        impl std::fmt::Display for $B {
+            fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
                 write!(f, "{}", &self.inner)
             }
         }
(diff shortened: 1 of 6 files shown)


jbaublitz:
I'm just catching up on this now. `Debug` and `Hash` are actually both considered "special" when it comes to procedural macros (the `#[derive(...)]` attributes). Both of these do not need to be fully qualified by their absolute path as neither of these derive statements introduces a dependency on a `use` outside of the macro. Because both of these are registered with the compiler as specific code generation keywords and procedural macros built by the core Rust team seem use absolute paths pretty consistently, this no longer becomes a symbol as much as the name of the code generation method to call at compile time. Because these methods are built in and not custom derives (see [here](https://doc.rust-lang.org/book/ch19-06-macros.html#how-to-write-a-custom-derive-macro) for more information on custom derives), they do not need a path when specified in a derive. `use` is only needed outside of the `#[derive(...)]` statement in custom derives when importing a procedural macro from another crate and even here, because it is outside of the generated code, will be necessary as a `use` and does not need to be avoided.

src/id_macros.rs
```diff
@@ -35,21 +35,21 @@ macro_rules! str_check {
 macro_rules! str_id {
     ($B:ident, $O:ident, $MAX:ident, $err_func:ident) => {
         /// The borrowed version of the DM identifier.
-        #[derive(Debug, PartialEq, Eq, Hash)]
+        #[derive(std::fmt::Debug, PartialEq, Eq, std::hash::Hash)]
```
jbaublitz: Please change these to `Debug` and `Hash` per my comment on the PR.

src/id_macros.rs
```diff
@@ -35,21 +35,21 @@ macro_rules! str_check {
...
         pub struct $B {
             inner: str,
         }
 
         /// The owned version of the DM identifier.
-        #[derive(Debug, Clone, PartialEq, Eq, Hash)]
+        #[derive(std::fmt::Debug, Clone, PartialEq, Eq, std::hash::Hash)]
```
jbaublitz: Please change these to Debug and Hash per my comment on the PR.

src/id_macros.rs
bgurney-rh: Resolved.

src/id_macros.rs
bgurney-rh: Resolved.

Issue #526, opened 2019-12-20
Fix beta clippy errors


mulkieran:
There are some new clippy in Rust 1.41.0.

PR #527, opened 2019-12-23
Remove tabs in doc comments

bgurney-rh:
Closes: #526


bgurney-rh:
I ran the CI "dm_test" after having executed "rustup default beta", and the results are successful.

PR #530, opened 2020-01-10
Run all tests for lowest supported toolchain as well as for stable

mulkieran:


mulkieran:
This will surely work, and then we can drop all Jenkins tests for devicemapper entirely.

mulkieran:
There was one of those intermittent ebusy errors that we get in the tests. We should obviously fix things there so that the tests are more robust.

PR #549, opened 2020-04-22
travis.yml: temporarily allow beta task to fail

bgurney-rh:
Related: stratis-storage/project#168


mulkieran:
I'm going to close this one. It turns out that devicemapper doesn't have any new beta errrors (this time), so we don't have to allow it. If the beta had come out tomorrow, as I expected it to, then this would have been merged.

PR #551, opened 2020-04-27
Increase recommended development toolchain to 1.43

bgurney-rh:
Related: stratis-storage/project#174


bgurney-rh:
I'm going to retry that CI job; it almost looked like a system problem occurred on the instance that ran the test.

Issue #485, opened 2019-07-16
Refactor parse_feature_args function call


waltdisgrace:
A slice of the vector containing the target parameter strings is being passed to the parse_feature_args function. This could lead to an out-of-bounds error that fails the program at runtime.

mulkieran:
@waltdisgrace Please generalize this issue so it covers the general case where the value for the number of feature args can be incorrect and lead to an out-of-bounds error. It describes the code in most, or possibly all, of the targets implemented.

mulkieran:
Some notes:

What you really want to look for are ```parse::<Type>``` lines, as with: https://github.com/stratis-storage/devicemapper-rs/blob/master/src/lineardev.rs#L438. Note that this in the method ```from_raw_table```. This is where the entire table is parsed from the test result returned by the ioctl. The method ```parse```, exists because ```LinearDevTargetParams``` implements ```FromStr```: https://github.com/stratis-storage/devicemapper-rs/blob/master/src/lineardev.rs#L362. This method invokes ```parse``` twice, one for ```FlakeyTargetParams``` and one for ```LinearTargetParams```. ```LinearTargetParams``` implements ```FromStr``` but is too simple to have features. ```FlakeyTargetParams``` implements ```FromStr```, and does have features. At https://github.com/stratis-storage/devicemapper-rs/blob/master/src/lineardev.rs#L302 the first argument parsed is the number of features args. The problems is that then we trust that number unhesitatingly. In particular, we use that number to select a slice from the remaining list of items. If that number is actually longer than the list of items, our code will crash. You can try that easily, with the Rust playground, using something like this:

```
fn main() {
    let v = vec![1, 2, 3, 4];
    let d = &v[2..5];
}
```

You will find that the same problem shows up in most but maybe not all of our implmentations of ```FromStr``` for other devices. For example:

https://github.com/stratis-storage/devicemapper-rs/blob/master/src/thinpooldev.rs#L92

PR #556, opened 2020-06-22
Check for correctness of user-inputted feature argument number

waltdisgrace:
Preliminary PR for #485


jbaublitz:
@waltdisgrace Ping me here once you've figured out the testing issues, happy to help if you get stuck.

src/lineardev.rs
```diff
@@ -918,6 +934,28 @@ mod tests {
...
+            .unwrap();
+        let expected = [FeatureArg::ErrorWrites, FeatureArg::DropWrites]
+            .iter()
+            .cloned()
+            .collect::<HashSet<_>>();
+        assert_eq!(result.feature_args, expected);
+    }
+
+    #[test]
+    #[should_panic]
```
mulkieran, line 950: What we're aiming for is for the test not to panic under these circumstances. So, you should remove the ```should_panic``` expectation on this test, and then fix what is causing the panic.

Note that if you run just the tests that you added, without the src code changes, they all pass, i.e., there's a panic during execution of this test. This is exactly what your code changes are supposed to fix.

waltdisgrace:
Rebased.

mulkieran:
@waltdisgrace Please post the error that you obtain from your newly added test with the _unmodified_ code.

waltdisgrace:
@mulkieran:

Here is the relevent snippet of the `make test` output:

```
failures:

---- lineardev::tests::test_flakey_incorrect_feature_args_input stdout ----
thread 'lineardev::tests::test_flakey_incorrect_feature_args_input' panicked at 'index 9 out of range for slice of length 8', src/lineardev.rs:303:18
stack backtrace:
   0: <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt
   1: core::fmt::write
   2: std::io::Write::write_fmt
   3: std::io::impls::<impl std::io::Write for alloc::boxed::Box<W>>::write_fmt
   4: std::panicking::default_hook::{{closure}}
   5: std::panicking::default_hook
   6: std::panicking::rust_panic_with_hook
   7: rust_begin_unwind
   8: core::panicking::panic_fmt
   9: core::slice::slice_index_len_fail
  10: <core::ops::range::Range<usize> as core::slice::SliceIndex<[T]>>::index
             at /builddir/build/BUILD/rustc-1.44.0-src/src/libcore/slice/mod.rs:2919
  11: core::slice::<impl core::ops::index::Index<I> for [T]>::index
             at /builddir/build/BUILD/rustc-1.44.0-src/src/libcore/slice/mod.rs:2732
  12: <alloc::vec::Vec<T> as core::ops::index::Index<I>>::index
             at /builddir/build/BUILD/rustc-1.44.0-src/src/liballoc/vec.rs:1945
  13: <devicemapper::lineardev::FlakeyTargetParams as core::str::FromStr>::from_str
             at src/lineardev.rs:303
  14: core::str::<impl str>::parse
             at /builddir/build/BUILD/rustc-1.44.0-src/src/libcore/str/mod.rs:4280
  15: devicemapper::lineardev::tests::test_flakey_incorrect_feature_args_input
             at src/lineardev.rs:923
  16: devicemapper::lineardev::tests::test_flakey_incorrect_feature_args_input::{{closure}}
             at src/lineardev.rs:922
  17: core::ops::function::FnOnce::call_once
             at /builddir/build/BUILD/rustc-1.44.0-src/src/libcore/ops/function.rs:232
  18: test::run_test::run_test_inner::{{closure}}
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

mulkieran:
@waltdisgrace: Ok! That failure looks right to me.

src/lineardev.rs
```diff
@@ -296,21 +296,38 @@ impl FromStr for FlakeyTargetParams {
         let up_interval = parse_value(vals[3], "up interval")?;
         let down_interval = parse_value(vals[4], "down interval")?;
 
-        let feature_args = if vals.len() == 5 {
-            vec![]
-        } else {
-            parse_feature_args(
+        if vals.len() == 5 {
```
mulkieran, line 299: Try to restructure the computation so that it calculates ```feature_args``` directly. Note that nothing else about the resulting ```FlakeyTargetParams``` object changes besides the value of the ```feature_args``` argument that is passed to the constructor.

So the computations should be more like:

```
let feature_args = <pretty complicated computation, returns an error if there is a problem>;
Ok(FlakeyTargetParams::new(..., feature_args))
```

src/lineardev.rs
```diff
@@ -296,38 +296,32 @@ impl FromStr for FlakeyTargetParams {
...
-            ))
+        let feature_args = if vals.len() == 5 {
+            vec![]
         } else if vals
             .as_slice()
             .get(5 + parse_value::<usize>(vals[5], "number of feature args")?)
             .is_some()
         {
-            let feature_args = parse_feature_args(
+            parse_feature_args(
```
mulkieran, line 306: This is moving in the correct direction. Note, though, that you've got another repeated computation where you parse the items at index 5 twice. See if you can find a way to parse that value only once in this block.

waltdisgrace:
Rebased

waltdisgrace:
@mulkieran Should the fixes for the other targets be done in separate PRs or the same PR as Flakey?

mulkieran:
> @mulkieran Should the fixes for the other targets be done in separate PRs or the same PR as Flakey?

@waltdisgrace You should make up a separate commit in this PR for ThinPoolDev and for each of the other target types that require the change.

waltdisgrace:
Rebased

src/cachedev.rs
```diff
@@ -97,19 +97,40 @@ impl FromStr for CacheTargetParams {
...
+
+        let num_feature_args = if vals.len() == 5 {
+            0
+        } else {
+            parse_value::<usize>(vals[5], "number of feature args")?
+        };
+
+        let feature_args: Vec<String> = if num_feature_args == 0 {
+            vec![]
+        } else {
```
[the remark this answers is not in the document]
mulkieran: Right after ```end_feature_args_index``` is calculated is where you should check that there are enough remaining values that you won't panic by accessing outside the array. You should write a test to verify that this works as well.

waltdisgrace:
I pushed a commit containing all of the changes you explicitly requested.

Regarding the test you provided: This test causes a panic at the assertion that the input results in an error, not at the source code. The name you chose for the test, "... too_few", implies that the input, "cache 42:42 42:43 42:44 16 2 writethrough passthrough default 0", has too few arguments. However, this is a valid input that has enough arguments. Am I missing something?

mulkieran:
> I pushed a commit containing all of the changes you explicitly requested.
> 
> Regarding the test you provided: This test causes a panic at the assertion that the input results in an error, not at the source code. The name you chose for the test, "... too_few", implies that the input, "cache 42:42 42:43 42:44 16 2 writethrough passthrough default 0", has too few arguments. However, this is a valid input that has enough arguments. Am I missing something?

Whoops! Yes, I had modified the test to try out some other things. Try this instead:
```
    #[test]
    fn test_cache_target_params_too_few() {
        let result = "cache 42:42 42:43 42:44 16 4 writethrough passthrough default 0"
            .parse::<CacheTargetParams>();
        assert_matches!(result, Err(DmError::Dm(ErrorEnum::Invalid, _)));
    }
```

waltdisgrace:
@mulkieran. That test looks reasonable to me and it passes when run against the source code changes of my most recently pushed commit. Adding your test to the PR now.

src/cachedev.rs
```diff
@@ -97,19 +97,40 @@ impl FromStr for CacheTargetParams {
...
+                .map(|x| (*x).to_string())
+                .collect()
+        };
 
         let policy = vals[end_feature_args_index].to_owned();
 
         let num_policy_args: usize =
             parse_value(vals[end_feature_args_index + 1], "number of policy args")?;
 
+        let num_expected_args = 8 + num_feature_args + num_policy_args;
```
mulkieran, line 128: Just drop this block for now. We should tackle this policy arg issue later.

src/cachedev.rs
```diff
@@ -97,19 +97,40 @@ impl FromStr for CacheTargetParams {
...
+            parse_value::<usize>(vals[5], "number of feature args")?
+        };
 
         let end_feature_args_index = 6 + num_feature_args;
-        let feature_args: Vec<String> = vals[6..end_feature_args_index]
-            .iter()
-            .map(|x| (*x).to_string())
-            .collect();
+
+        if vals.len() <= end_feature_args_index + 1 {
```
mulkieran, line 109: Just make this ```if vals.len() < end_feature_args_index``` and then make the error message specific to too few feature_args. That change means we could still fail on policy missing, but I want to leave that until later, also.

src/cachedev.rs
```diff
@@ -97,19 +97,40 @@ impl FromStr for CacheTargetParams {
...
+            let err_msg = format!("Only {} arguments provided, expected more", vals.len());
```
mulkieran, line 110: Try to make this error message as specific as the ones for flakey and thinpool and roughly the same format.

waltdisgrace: I did not make this error message as specific as the ones for flakey and thinpool, because:

Here, there exist (mandatory) arguments which come after the feature arguments, while in flakey and thinpool, there does not exist such arguments. Here, said arguments, i.e. the arguments which come after the feature arguments, can be interpreted as feature arguments if the user provides a too-large number-of-feature-arguments argument. Given the aforementioned information, we can assert less about the user's mistake here than we can in the situation of flakey or thinpool.

Let's examine the following example input:
"cache 42:42 42:43 42:44 16 3 writethrough passthrough default 0"

In this example, the user could've made one of two mistakes:

- Possibility 1: The user could've intended that there be two feature arguments, "writethrough" and "passthrough", thereby mistakenly inputting "3" as the number-of-feature-arguments argument.
- Possibility 2: The user could've intended that there be three feature arguments, "writethrough," "passthrough," and "default," thereby intentionally inputting "3" as the number-of-feature-arguments argument, _however_ mistakenly believing that "default" is a valid feature argument and also mistakenly neglecting to input a policy argument.

If only Possibility 1 was possible, then our certainty level about the user's error would increase. Given the existence of Possibility 2, we remain uncertain about the user's intention and mistake(s).

Stating "Expected 3 feature arguments, found 2" would not be helpful to a user who inputted 3, not fully valid feature arguments. In this case, stating that "default" is not a valid feature argument could be a possibility, though we would have to parse the feature arguments for their validity in order for that to work.

waltdisgrace: When choosing a specificity level of the error message returned, another relevant example to consider is "cache 42:42 42:43 42:44 16 3 writethrough passthrough metadata2 0". This is due to similar reasons as above, though in this case, the user inputted 3 correct feature arguments, and neglected to enter a policy argument.

mulkieran: @waltdisgrace You're correct about the many different types of errors that a programmer might introduce. But our goal is not to understand and explain all the errors, because we actually can't w/out adding a lot more infrastructure. Our goal is to make sure that devicemapper code doesn't panic if the stuff that it has to parse is incorrectly formatted, and to make the code as readable and consistent as possible.

src/cachedev.rs
waltdisgrace: To avoid an out-of-bounds panic on test inputs that are only incorrect because of an incorrect feature argument number, this condition was chosen. Changing the condition to ```if vals.len() < end_feature_args_index``` would make sense for test inputs that are incorrect for the additional reason of being void of mandatory policy args, e.g. "cache 42:42 42:43 42:44 16 4 writethrough".

mulkieran:
Unlikely to manage to get to this before next release.

Issue #557, opened 2020-06-23
Travis devicemapper-rs tests failing regularly with EBUSY


jbaublitz:
@mulkieran Travis tests were failing for @waltdisgrace pretty regularly with EBUSY on different loopback device tests every time. We seem to have gotten #556 to pass by rerunning it enough times but we should potentially look into what's going on here. I don't think this has anything to do with the code changes, so we may have to think about how we're testing in devicemapper-rs.

mulkieran:
Yep, I think this was always a bit of a problem, and I think it's just been happening more and more often. We will have to find a better way, maybe the delay and wait for a particular thing to be true technique will work here, too.

mulkieran:
We need to start taking this seriously, it is such a pain.

mulkieran:
Pushing to September in hopes that #568 will keep the problem from being annoying for a while.

mulkieran:
yeah, that sure worked. Let's not bother to think about it unless it starts bothering us again.

PR #568, opened 2020-07-22
Just remove troublesome suspend/resume testing in test_basic

mulkieran:
Related: #557


mulkieran:
Note that this is not a fix to our general EBUSY testing problem, but I think it will be shown to be an annoyance remover.

PR #577, opened 2020-09-04
Use unknown_clippy_lint allow to allow a beta-only-lint on the stable branch

mulkieran:


mulkieran:
This suggestion will improve things a bit, as we can easily know that a ```unknown_clippy_lint``` annotation indicates something that should be removed at some time. Suggested here: https://github.com/rust-lang/rust-clippy/issues/6002#issuecomment-686877681

bgurney-rh (approved):

PR #592, opened 2020-12-10
Implement AsRawFd for the devicemapper context

mulkieran:


mulkieran:
I think that the principle reason for the Eventable trait in stratisd is that DM did not implement AsRawFd. I'm hoping to get rid of the Eventable trait entirely in some future time. This is one part of the plan. Also, implementing AsRawFd for DM seems like the right thing to do anyway.

jbaublitz (approved):

PR #598, opened 2021-01-08
Migrate Bytes to u128 internally

jbaublitz:
Supercedes #515


mulkieran:
"This should probably be a minor version bump as it breaks the API for the return type of arithmetic operations." Avoiding putting this in PR message proper. I agree, we'll do a minor version bump.

mulkieran (approved):

PR #601, opened 2021-01-09
Exclude administrative files from crate

mulkieran:
Related https://github.com/stratis-storage/project/issues/256


jbaublitz (approved):
Just a note that I typically use a whitelist instead of a blacklist but I'm fine with this. Cargo has syntax to only include the Rust files, license, and any other files you'd like.

mulkieran:
Ok! If you really don't care I'ld prefer to stick with exclude.

mulkieran:
rebased.

PR #603, opened 2021-02-01
Change MAX_RECOMMENDED_METADATA_SIZE to MAX_METADATA_SIZE

mulkieran:
Related https://github.com/stratis-storage/stratisd/issues/1495


mulkieran:
I don't want to expose the constant as part of the devicemapper interface as there's a question about the appropriate units. The constant is in MetaBlocks, which works right now because meta block sizes are always constant. Should that change, it might be better to expose the size in Sectors or even Bytes.

bgurney-rh (approved):

PR #609, opened 2021-03-18
Add CI task to build with minimal dependencies

mulkieran:


mulkieran:
Must spell "minimal" correctly...

mulkieran:
Let's see how stratisd goes. It is always the most difficult.

mulkieran:
We should use the script we use in stratisd rather than what we see here, closing.

PR #620, opened 2021-04-23
github actions: test nightly

bgurney-rh:


bgurney-rh:
This is all set; I was able to find an issue with the nightly.yml file, which is fixed in #621

PR #624, opened 2021-04-28
Cleanup

### LINKED ISSUE #525: Remove expect calls on Option result of slice_to_null
These assume that the array of bytes being examined will always have a "\0" character somewhere in it. This is very likely, but absolutely not guaranteed; it would be a shame just to crash if this proved to be false.

### LINKED ISSUE #623: Use bindgen to dynamically generate low-level ioctl bindings
Currently, our file ```dm_ioctl.rs``` contains low-level bindings to the ioctl data structures that were generated at some prior time using bindgen as well as some additional code that was added later.

We would be better off using bindgen in the build process to dynamically generate the bindings and adding a more abstract layer separately as necessary.

jbaublitz:
Closes #623
Closes #525


jbaublitz:
@mulkieran Just putting up for preliminary review. What do you think so far?

mulkieran (commented):
Generally speaking, looks good to me. Have you considered generating the header file, in this case, ```dm-ioctl.h``` in the build.rs script?

jbaublitz:
Per our discussion, let's leave it as is given that the header is more configuration than code.

mulkieran:
Got the feeling that this is just about done today, so putting it in progress.

jbaublitz:
Okay, this seems to be ready for final review!

build.rs
```diff
@@ -0,0 +1,22 @@
...
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+use std::{env::var, path::PathBuf};
+
+use bindgen::Builder;
+
+fn main() {
+    let bindings = Builder::default()
+        .header("dm-ioctl.h")
```
flxo, line 11: `bindgen::Builder` has a nice [header_contents](https://docs.rs/bindgen/0.58.1/bindgen/struct.Builder.html#method.header_contents) fn that is perfect to avoid those small "just include something" files in the tree.

mulkieran: thx for the tip!

mulkieran: @jbaublitz Regarding using using ```header_contents()``` in this project; that would be fine w/ me, but also using the header file is fine w/ me.

mulkieran (changes requested):
Looks good. Please grab commit https://github.com/stratis-storage/devicemapper-rs/pull/632/commits/699fba8fcf34959d67f110a037202ea3d63f0d65 if that looks good, and this should be ready.

jbaublitz:
@mulkieran Okay change applied!

mulkieran (changes requested):
Looks good! Please close this and open a PR for final review.

PR #630, opened 2021-06-10
Add compare_fedora_versions tests

bgurney-rh:
Related stratis-storage/project#307


mulkieran (changes requested):
The structure is good, but a bunch of the dependencies, e.g., dbus-devel, don't apply to devicemapper-rs. Try to drop as many of those as possible.

PR #638, opened 2021-06-22
Rust 1.53 clippy updates

bgurney-rh:


src/cachedev.rs
```diff
@@ -855,7 +855,7 @@ mod tests {
 
                 assert_eq!(status.metadata_mode, CacheDevMetadataMode::Good);
 
-                assert_eq!(status.needs_check, false);
+                assert!(status.needs_check, "{}", false);
```
mulkieran, line 858: There's a general pattern to follow here:

```assert_eq!(<expression>, false)``` -> ```assert!(!<expression>)```.

That should be quite adequate here. You _can_ use the two additional fields for a panic message  if you want to, but that's more work than necessary.

mulkieran (changes requested):

src/cachedev.rs
bgurney-rh: Resolved.

bgurney-rh:
I'm going to create a new version of this PR; still with 2 commits, but cleaned up.

mulkieran (changes requested):
Looks ready! Please close and open a new PR for final review.

Issue #640, opened 2021-06-23
Build is broken for target aarch64-linux-android


flxo:
The bionic interface slightly differs from glibc and others. The build is broken due to some type conflict:

```
error[E0605]: non-primitive cast: `&[u8; 129]` as `&[i8]`
  --> src/core/deviceinfo.rs:44:35
   |
44 |         let uuid = str_from_c_str(&ioctl.uuid as &[i8]).ok_or_else(|| {
   |                                   ^^^^^^^^^^^^^^^^^^^^ an `as` expression can only be used to convert between primitive types or to coerce to a specific trait object

error[E0605]: non-primitive cast: `&[u8; 128]` as `&[i8]`
  --> src/core/deviceinfo.rs:71:32
   |
71 |                 str_from_c_str(&ioctl.name as &[i8])
   |                                ^^^^^^^^^^^^^^^^^^^^ an `as` expression can only be used to convert between primitive types or to coerce to a specific trait object

error[E0308]: mismatched types
  --> src/core/dm.rs:98:59
   |
98 |         let _ = name.as_bytes().read(mut_slice_from_c_str(&mut hdr.name))?;
   |                                                           ^^^^^^^^^^^^^ expected slice `[i8]`, found array `[u8; 128]`
   |
   = note: expected mutable reference `&mut [i8]`
              found mutable reference `&mut [u8; 128]`

error[E0308]: mismatched types
   --> src/core/dm.rs:103:59
    |
103 |         let _ = uuid.as_bytes().read(mut_slice_from_c_str(&mut hdr.uuid))?;
    |                                                           ^^^^^^^^^^^^^ expected slice `[i8]`, found array `[u8; 129]`
    |
    = note: expected mutable reference `&mut [i8]`
               found mutable reference `&mut [u8; 129]`

error[E0308]: mismatched types
   --> src/core/dm.rs:481:44
    |
481 |             let dst = mut_slice_from_c_str(&mut targ.target_type);
    |                                            ^^^^^^^^^^^^^^^^^^^^^ expected slice `[i8]`, found array `[u8; 16]`
    |
    = note: expected mutable reference `&mut [i8]`
               found mutable reference `&mut [u8; 16]`

error[E0308]: mismatched types
   --> src/core/dm.rs:586:50
    |
586 |                 let target_type = str_from_c_str(&targ.target_type)
    |                                                  ^^^^^^^^^^^^^^^^^ expected slice `[i8]`, found array `[u8; 16]`
[...]

PR #641, opened 2021-06-23
Add a job to do an architecture specific build

mulkieran:
Related #640

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -78,7 +78,7 @@ jobs:
       - uses: actions-rs/cargo@v1
         with:
           use-cross: true
-          command: build
+          command: check
           args: --target ${{ matrix.target }}
 
   checks-with-ci-repo:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -75,11 +75,14 @@ jobs:
         run: |
           sudo apt-get -q update
           sudo apt-get -y install libdevmapper-dev
-      - uses: actions-rs/cargo@v1
-        with:
-          use-cross: true
-          command: check
-          args: --target ${{ matrix.target }}
+      - name: Install core/std for ${{ matrix.target }}
+        run: >
+          rustup target add
+          --toolchain ${{ matrix.toolchain }}
+          ${{ matrix.target }}
+      - name: Run 'cargo check' on the specified target
+        run: cargo check --target ${{ matrix.target }}
+
 
   checks-with-ci-repo:
     strategy:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -56,6 +56,31 @@ jobs:
       - name: Test ${{ matrix.task }} on ${{ matrix.toolchain }} toolchain
         run: ${{ matrix.task }}
 
+  architecture-specific-checks:
+    strategy:
+      matrix:
+        include:
+          - components: cargo
+            toolchain: 1.52.1
+            target: aarch64-linux-android
+    runs-on: ubuntu-18.04
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
+        with:
+          toolchain: ${{ matrix.toolchain }}
+          target: ${{ matrix.target }}
+          override: true
+      - name: Install dependencies
+        run: |
+          sudo apt-get -q update
+          sudo apt-get -y install libdevmapper-dev
+      - uses: actions-rs/cargo@v1
+        with:
+          use-cross: true
+          command: build
+          args: --target ${{ matrix.target }}
+
   checks-with-ci-repo:
     strategy:
       matrix:


flxo:
@mulkieran I made the experience that (at least for `aarch64-linux-android`) there's no way around [cross](https://github.com/rust-embedded/cross). Installing the C (NDK) toolchain isn't possible via all those package managers and a mess to do manually. The rust actions can directly invoke `cross` by setting the `with-cross` flag ([example](https://github.com/esrlabs/northstar/blob/e09fdb1ebfb4bf6306e7c691f62d81d289a26b63/.github/workflows/cross.yml#L29-L34)).

jbaublitz:
@flxo My thought was going to be using docker containers. Do you have any objections to this? Generally this would not allow us to test devicemapper runs given that the kernel would still be Ubuntu but we could at least check that it compiles.

flxo:
@jbaublitz I think that using `Docker` with `cross` is the only feasible way. When dealing with libc headers there are so many differences between the architectures.
I did a quick hack to check #643 . It [builds](https://github.com/flxo/devicemapper-rs/runs/2901801730?check_suite_focus=true) for `aarch64-linux-android` and uses a Docker image with an updated NDK.

The NDK use in the original cross docker image for `aarch64-linux-android` is outdated. Overriding the Docker images for cross is done via a [Cross.toml](https://github.com/flxo/devicemapper-rs/blob/master/Cross.toml).

Of course you can start brewing you own images but this is probably no fun. We started of the cross images and extended them.

flxo:
I want to summarise what I know about the topic:

# `aarch64-unknown-linux-android` and `aarch64-unknown-linux-musl`

The `cross` docker images lacks `libclang-dev` which is needed for `bindgen`. A `Dockerfile` for those two architctures would look like

```docker
FROM rustembedded/cross:aarch64-unknown-linux-gnu
RUN apt update && apt install --assume-yes libclang-dev
```

The `devicemapper-rs` crate would need a `Cross.toml` to tell cross that is shall use the images from somewhere:

Cross.toml:
```
[target.aarch64-unknown-linux-gnu]
image = "stratis/aarch64-unknown-linux-gnu:latest"

[target.aarch64-unknown-linux-musl]
image = "stratis-aarch64-unknown-linux-musl:latest"
```

Additionally the clang invocation from `build.rs` needs an additional include path:

```rust
    let target = std::env::var("TARGET").unwrap();

    // Need some extra include path for the cross images for aarch64 gnu and musl
    let extra_arg = match target.as_str() {
        "aarch64-unknown-linux-gnu" => "-I/usr/aarch64-linux-gnu/include",
        "aarch64-unknown-linux-musl" => "-I/usr/local/aarch64-linux-musl/include",
        _ => "",
    }

    let bindings = Builder::default()
        .header("dm-ioctl.h")
        .derive_debug(true)
        .derive_default(true)
        .clang_arg(extra_arg)
        .generate()
        .expect("Could not generate bindings");
```

The two things above satisfy the `bindgen` invocation as part of `cross check --target aarch64-unkown-linux-gnu` and `cross check --target aarch64-unkown-linux-musl`

## Compilation

The kernel API differs for those two architectures and some type conversions similar to #643 are needed. The current master + #643 fails with:

```
northstar ~/devicemapper-rs ‹pr-android-build*› cross check --target aarch64-unknown-linux-gnu
   Compiling devicemapper v0.29.1 (/project)
error[E0425]: cannot find value `DM_DEV_ARM_POLL_CMD` in module `dmi`
   --> src/core/dm.rs:766:28
    |
766 |         self.do_ioctl(dmi::DM_DEV_ARM_POLL_CMD as u8, &mut hdr, None)?;
    |                            ^^^^^^^^^^^^^^^^^^^ help: a constant with a similar name exists: `DM_DEV_REMOVE_CMD`
    | 
   ::: /target/aarch64-unknown-linux-gnu/debug/build/devicemapper-dcebdc152793c932/out/bindings.rs:584:1
    |
584 | pub const DM_DEV_REMOVE_CMD: ::std::os::raw::c_uint = 4;
    | -------------------------------------------------------- similarly named constant `DM_DEV_REMOVE_CMD` defined here

error[E0605]: non-primitive cast: `&[u8; 129]` as `&[i8]`
  --> src/core/deviceinfo.rs:48:35
   |
48 |         let uuid = str_from_c_str(&ioctl.uuid as &[i8]);
   |                                   ^^^^^^^^^^^^^^^^^^^^ an `as` expression can only be used to convert between primitive types or to coerce to a specific trait object

error[E0605]: non-primitive cast: `&[u8; 128]` as `&[i8]`
  --> src/core/deviceinfo.rs:64:35
   |
64 |         let name = str_from_c_str(&ioctl.name as &[i8]);
   |                                   ^^^^^^^^^^^^^^^^^^^^ an `as` expression can only be used to convert between primitive types or to coerce to a specific trait object

error[E0308]: mismatched types
  --> src/core/dm.rs:99:45
   |
99 |         let hdr_name = mut_slice_from_c_str(&mut hdr.name);
   |                                             ^^^^^^^^^^^^^ expected slice `[i8]`, found array `[u8; 128]`
   |
   = note: expected mutable reference `&mut [i8]`
              found mutable reference `&mut [u8; 128]`

error[E0308]: mismatched types
   --> src/core/dm.rs:109:45
    |
109 |         let hdr_uuid = mut_slice_from_c_str(&mut hdr.uuid);
    |                                             ^^^^^^^^^^^^^ expected slice `[i8]`, found array `[u8; 129]`
    |
    = note: expected mutable reference `&mut [i8]`
               found mutable reference `&mut [u8; 129]`

error[E0308]: mismatched types
   --> src/core/dm.rs:492:44
    |
492 |             let dst = mut_slice_from_c_str(&mut targ.target_type);
    |                                            ^^^^^^^^^^^^^^^^^^^^^ expected slice `[i8]`, found array `[u8; 16]`
    |
    = note: expected mutable reference `&mut [i8]`
               found mutable reference `&mut [u8; 16]`

error[E0308]: mismatched types
   --> src/core/dm.rs:601:50
    |
601 |                 let target_type = str_from_c_str(&targ.target_type);
    |                                                  ^^^^^^^^^^^^^^^^^ expected slice `[i8]`, found array `[u8; 16]`
    |
    = note: expected reference `&[i8]`
               found reference `&[u8; 16]`

error[E0605]: non-primitive cast: `&[u8; 128]` as `&[i8]`
  --> src/core/dm_ioctl.rs:32:35
   |
32 |         let name = str_from_c_str(&self.name as &[i8]);
   |                                   ^^^^^^^^^^^^^^^^^^^ an `as` expression can only be used to convert between primitive types or to coerce to a specific trait object

error[E0605]: non-primitive cast: `&[u8; 129]` as `&[i8]`
  --> src/core/dm_ioctl.rs:37:35
   |
37 |         let uuid = str_from_c_str(&self.uuid as &[i8]);
   |                                   ^^^^^^^^^^^^^^^^^^^ an `as` expression can only be used to convert between primitive types or to coerce to a specific trait object

error[E0605]: non-primitive cast: `&[u8; 7]` as `&[i8]`
  --> src/core/dm_ioctl.rs:42:35
   |
42 |         let data = str_from_c_str(&self.data as &[i8]);
   |                                   ^^^^^^^^^^^^^^^^^^^ an `as` expression can only be used to convert between primitive types or to coerce to a specific trait object

error: aborting due to 10 previous errors

Some errors have detailed explanations: E0308, E0425, E0605.
For more information about an error, try `rustc --explain E0308`.
error: could not compile `devicemapper`

To learn more, run the command again with --verbose.
```

Should be fixable. 

# `aarch64-linux-android`

The NDK in the [upstream cross image](https://github.com/rust-embedded/cross/blob/7405f27f0a6d58df736dbb93fa7af5bb6c37302c/docker/android-ndk.sh#L6) for `aarch64-linux-android` quite old. Back in the days I decided to replace the NDK in the cross image with a recent one that matches my target platform version, so I cannot tell if it's possible to use the upstream cross image.

For devicemapper-rs the target should be to use the upstream cross image which is probably the most common way someone would try to compile this crate for android (if anybody beside me is crazy enough to do so ;-)).

The Android toolchain setup in the Dockerfile for `aarch64-linux-android` with the wrappers around clang that set a proper sys root etc do not require any additions to the `build.rs` and for the targets listed above.

hope this helps!

mulkieran:
Superceded by #647

PR #649, opened 2021-06-25
Add CI for musl platforms

### LINKED ISSUE #648: Add CI for musl platforms
We have previously broken compatibility for musl so we should add CI checks to detect breakage before it's merged.

jbaublitz:
Closes #648

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -56,6 +56,50 @@ jobs:
       - name: Test ${{ matrix.task }} on ${{ matrix.toolchain }} toolchain
         run: ${{ matrix.task }}
 
+  musl-checks:
+    strategy:
+      matrix:
+        include:
+          # MANDATORY CHECKS USING CURRENT DEVELOPMENT COMPILER
+          - components: rustfmt
+            toolchain: 1.53.0
+            task: make -f Makefile travis_fmt
+          - components: clippy
+            toolchain: 1.53.0
+            task: make -f Makefile clippy
+          # MANDATORY TESTING USING LOWEST SUPPORTED COMPILER
+          - components: cargo
+            toolchain: 1.49.0
+            task: >
+              sudo PATH=$GITHUB_WORKSPACE/.cargo/bin:$PATH
+              make -f Makefile sudo_test
+          # MANDATORY TESTING ON STABLE
+          - components: cargo
+            toolchain: stable
+            task: make -f Makefile build
+          - components: cargo
+            toolchain: stable
+            task: >
+              sudo PATH=$GITHUB_WORKSPACE/.cargo/bin:$PATH
+              make -f Makefile sudo_test
+          - components: cargo
+            toolchain: stable
+            task: make -f Makefile docs
+    runs-on: ubuntu-18.04
+    container:
+      image: alpine
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions-rs/toolchain@v1
+        with:
+          components: ${{ matrix.components }}
+          toolchain: ${{ matrix.toolchain }}
+          override: true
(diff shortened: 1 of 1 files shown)


jbaublitz:
Re-ran tests and they passed. Putting up for review.

mulkieran (changes requested):
Looks good. Please close this PR and open a new one for final review.

.github/workflows/main.yml
```diff
@@ -56,6 +56,17 @@ jobs:
...
         run: ${{ matrix.task }}
 
+  musl-checks:
+    runs-on: ubuntu-18.04
+    container:
+      image: alpine
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install dependencies
+        run: apk add cargo clang linux-headers make rust
```
flxo, line 66: nitpick: For `bindgen` we need `lib clang*.so` which is part of `libclang-dev`. `clang` itself is not needed so installing `libclang-dev` instead of `clang` safes a couple of bytes....but totally neglect able. 
Same in #649.

flxo: `rust`: What's the reason to use the packaged apk instead of `actions-rs/toolchain@v1` and pull the latest stable?

.github/workflows/main.yml
```diff
@@ -56,6 +56,17 @@ jobs:
...
+      - name: Test build on Alpine Linux
```
flxo, line 67: `make build` runs `RUSTFLAGS="${DENY}" cargo build`. Think the `name` is outdated and there's no Alpine image involved.

.github/workflows/main.yml
mulkieran: I'ld be happy w/ using toolchain Action and pulling in current development toolchain, also.

jbaublitz: @flxo The reason for this is that rustc does not end up in the `PATH` if installed via the toolchain action. Currently there doesn't seem to be a way to fix that as Github Actions appears to have deprecated certain ways of adding an executable to the path.

.github/workflows/main.yml
jbaublitz: What do you mean? The build is happening on an Alpine Linux image.

.github/workflows/main.yml
flxo: @jbaublitz Ah ok. It's just usable by the `cargo` actions then. Didn't know that.

.github/workflows/main.yml
flxo: I'm confused. There's a `runs-on: ubuntu-18.04` right after `musl-checks`. The `Makefile` just calls `cargo` which must be the rust toolchain isnstalled via `apt`.

.github/workflows/main.yml
```diff
@@ -56,6 +56,17 @@ jobs:
       - name: Test ${{ matrix.task }} on ${{ matrix.toolchain }} toolchain
         run: ${{ matrix.task }}
 
+  musl-checks:
+    runs-on: ubuntu-18.04
+    container:
+      image: alpine
```
jbaublitz, line 62: @flxo This is the definition of the container that the build is running on. Ubuntu is just the host system because that's all that Github Actions provides.

flxo: @jbaublitz Thanks for the clarification!

jbaublitz:
@mulkieran I'm adding one additional change to this PR. Once CI passes, I'll open one for final review.

mulkieran:
> @mulkieran I'm adding one additional change to this PR. Once CI passes, I'll open one for final review.

@jbaublitz Ok. Test looks good.

PR #671, opened 2021-07-28
Remove error-chain dependency in source

mulkieran:


mulkieran:
tests in stratisd are similar, and I'll expect to drop the dependency on error_chain there in roughly the same way.

jbaublitz (approved):
This all looks good to me!

PR #677, opened 2021-07-31
Update nix to 0.22

remilauzier:


mulkieran:
thx! Much appreciated. This is what I should have expected, I think.

mulkieran (changes requested):
Please also bump version of libc to 0.2.98. nix 0.22.0 requires at least 0.2.98, so we might as well not specify anything less in Cargo.toml.

flxo:
> Please also bump version of libc to 0.2.98. nix 0.22.0 requires at least 0.2.98, so we might as well not specify anything less in Cargo.toml.

`nix` reexports it's `libc` dependency. Maybe remove the explicit `libc` dependency and use `nix::libc`? This is less noise in the cargo manifest and you never get out of sync.

mulkieran:
> > Please also bump version of libc to 0.2.98. nix 0.22.0 requires at least 0.2.98, so we might as well not specify anything less in Cargo.toml.
> 
> `nix` reexports it's `libc` dependency. Maybe remove the explicit `libc` dependency and use `nix::libc`? This is less noise in the cargo manifest and you never get out of sync.

@flxo thx, that seems like it's a good idea, since I imagine nix must always bump its required version of libc. We'll consider that for future work.

mulkieran (approved):

PR #680, opened 2021-08-07
Fix some warnings

remilauzier:


mulkieran:
@remilauzier What tool+version is giving you these warnings?

remilauzier:
It's rust_2018_idioms with rust stable. Hidden anonymous lifetime are deprecated and will be removed in the 2021 version. The remove u32 is IntelliJ IDEA. Using std:u32 are deprecated too. Using directly u32 is the new way.

mulkieran:
> It's rust_2018_idioms with rust stable. Hidden anonymous lifetime are deprecated and will be removed in the 2021 version. The remove u32 is IntelliJ IDEA. Using std:u32 are deprecated too. Using directly u32 is the new way.

Ok! In our Makefile we have a variable ```RUST_2018_IDIOMS```. Could you add the other lints in that category that this PR fixes to that list?

mulkieran:
@remilauzier I noticed that you edited ```lib.rs``` w/ a bunch of ```deny``` directives. Is that so that Intellij IDE and possibly other IDEs can read that info and act appropriately? I use a text editor for all my development, and haven't been following the evolution of IDEs in the last couple years. If that is really the case, then we might consider moving our management of compiler and clippy errors out of the Makefile and into the Rust source entirely. Please let me know, thx.

remilauzier:
It's been a few year that rustc and clippy read directly deny warning from the source file. That permit everyone to know what to do or not easily. It's really make to not depend on a single IDE. You can even put a ``#[allow(something)]`` directly for a single function if you want. You can even tell clippy to not let you see some warning if you put the ``msrv`` in ``clippy.toml``. Doing it in ``lib.rs`` or ``main.rs`` permit it to apply for all the project. I must said it's a first for me to have these in a makefile.

mulkieran:
@remilauzier This project has been around since before the "?" operator existed and when clippy could only be run on nightly! But, on reading around a bit, it looks like there are benefits to keeping these directives in a Makefile. Here's one pretty good reason: https://www.reddit.com/r/rust/comments/f5xpib/psa_denywarnings_is_actively_harmful/. And another: https://github.com/rust-lang/cargo/issues/5998. We already use a clippy.toml to set the msrv....that seems to have come in handy. But we'll stick with setting those clippy and compiler warning directives in the Makefile, for the foreseeable future.

mulkieran (changes requested):
Ok, thx for you answer. After some thought, I've decided to go w/ my original review. At some point, though, I'll probably try to move the two remaining directives in ```lib.rs``` into the Makefile as well.

mulkieran (approved):

jbaublitz (approved):

PR #727, opened 2021-12-09
trigger udev rules processing when adding linear devices

baloo:
Without this commit, added devices will not appear in the /dev/disk directory

### DIFF
--- a/src/lineardev.rs
+++ b/src/lineardev.rs
@@ -5,7 +5,7 @@
 use std::{collections::HashSet, fmt, path::PathBuf, str::FromStr};
 
 use crate::{
-    core::{DevId, Device, DeviceInfo, DmFlags, DmName, DmOptions, DmUuid, DM},
+    core::{DevId, Device, DeviceInfo, DmCookie, DmFlags, DmName, DmOptions, DmUuid, DM},
     result::{DmError, DmResult, ErrorEnum},
     shared::{
         device_create, device_exists, device_match, parse_device, parse_value, DmDevice,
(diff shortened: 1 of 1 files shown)


mulkieran:
@baloo Thanks for the PR! If I understand you correctly, you believe that when creating a linear device the ```DM_UDEV_PRIMARY_SOURCE_FLAG``` should _always_ be set?

baloo:
From what I can tell, that's is the behavior of dmsetup, and I can tell that without `PRIMARY_SOURCE_FLAG` udev won't process any of the rules to add the `/dev/disk/` symlinks.

I can try to make that optional, but I don't see a reason to do so.

baloo:
```
/*
 * DM_UDEV_PRIMARY_SOURCE_FLAG is automatically appended by
 * libdevmapper for all ioctls generating udev uevents. Once used in
 * udev rules, we know if this is a real "primary sourced" event or not.
 * We need to distinguish real events originated in libdevmapper from
 * any spurious events to gather all missing information (e.g. events
 * generated as a result of "udevadm trigger" command or as a result
 * of the "watch" udev rule).
 */
#define DM_UDEV_PRIMARY_SOURCE_FLAG 0x0040
```
This is the documentation for `PRIMARY_SOURCE_FLAG` and default udev rules will exit immediately if this flag is not set.

baloo:
Test failure is weird, I run those tests in a vm, the first run will hit similar issue, but subsequent run will run fine.

Restarting the vm will make it reproducible.

mulkieran:
@baloo Sorry for the delay! Your PR brought up some larger questions and we had to do a bit of research. I've put up a proposal[1] describing what I think is the best course of action and why I'm proposing it. Please feel free to comment there.

[1] ttps://github.com/stratis-storage/project/discussions/389

mulkieran:
> Hey @mulkieran thank you for that! I think the proposal makes sense, I don't think the breaking change is too much of an issue, it's an easy fix and I think overall rust devs are used to it.

Thanks, @baloo! We expect to go ahead w/ the proposed change shortly, but will be cutting a patch release of devicemapper-rs first.

mulkieran:
@baloo Sorry this took so long. We had confidence in the correctness of your change but found that we were nonetheless causing stratisd to behave incorrectly when it was included which caused us to hesitate. We believe that PR https://github.com/stratis-storage/devicemapper-rs/pull/774 will do the trick, however.

baloo:
it's all good! :)
Thanks for following up on this. I only tested that in a lab environment in an initrd. Unlikely to hit any concurrency issues there. Your solution looks better.

Thanks for the hard work.

PR #731, opened 2022-01-12
Default to thin pool queueing and allow changing behavior

jbaublitz:
Related to stratis-storage/stratisd#2814

### DIFF
--- a/src/thinpooldev.rs
+++ b/src/thinpooldev.rs
@@ -476,6 +476,7 @@ impl ThinPoolDev {
     /// on the metadata device. If the metadata is corrupted, subsequent
     /// errors will result, so it is expected that the metadata is
     /// well-formed and consistent with the data on the data device.
+    #[allow(clippy::too_many_arguments)]
     pub fn setup(
         dm: &DM,
         name: &DmName,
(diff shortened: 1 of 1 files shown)


jbaublitz:
Blocked by #737

src/thinpooldev.rs
```diff
@@ -508,6 +511,32 @@ impl ThinPoolDev {
...
+    /// where the thin-pool-specific string has the format:
+    /// <meta maj:min> <data maj:min> <block size> <low water mark>
+    /// There is exactly one entry in the table.
+    fn gen_table(
+        meta: &LinearDev,
+        data: &LinearDev,
+        data_block_size: Sectors,
+        low_water_mark: DataBlocks,
+        feature_args: Vec<String>,
+    ) -> ThinPoolDevTargetTable {
```
mulkieran, line 526: Should probably have ```gen_default_table``` invoke ```gen_table```.

mulkieran (changes requested):
Code looks good, one request. The idea of toggling ```error_if_no_space``` is a little odd to me, I would have felt more confident with methods to set and unset it (which would have been pretty minimalist). But, go w/ toggle if you're sure that's what you want. I already looked at the source in stratisd and I understand the usage.

jbaublitz:
> Code looks good, one request. The idea of toggling error_if_no_space is a little odd to me, I would have felt more confident with methods to set and unset it (which would have been pretty minimalist). But, go w/ toggle if you're sure that's what you want. I already looked at the source in stratisd and I understand the usage.

You may have a point here mainly because I think it makes a lot of sense from the point of view of stratisd as a consumer, but maybe less sense as a general solution in a library. Let me play around with this and see what the alternative looks like.

src/thinpooldev.rs
jbaublitz: Resolved!

mulkieran (approved):

mulkieran:
@jbaublitz Plz squash and merge Thursday.

PR #732, opened 2022-01-12
Increase version of ubuntu virtual environment

mulkieran:


mulkieran:
Not quite trivial, but it gets our CI back in business, so merging now.

PR #739, opened 2022-01-19
Eliminate some expect invocations

### LINKED ISSUE #434: devicemapper should dereference non-null pointers directly instead of using as_ref

mulkieran:
Closes #434

### DIFF
--- a/src/core/dm.rs
+++ b/src/core/dm.rs
@@ -143,11 +143,7 @@ impl DM {
         // zero out the rest
         let cap = v.capacity();
         v.resize(cap, 0);
-        let mut hdr = unsafe {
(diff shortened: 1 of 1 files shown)


jbaublitz (approved):

mulkieran:
rebased.

PR #741, opened 2022-01-19
Pass DM_UDEV_PRIMARY_SOURCE_FLAG with every ioctl

### LINKED ISSUE #740: Set DM_UDEV_PRIMARY_SOURCE_FLAG in ioctl
Extend device creation methods so that they are parameterized on a ```create_options``` parameter.

Related https://github.com/stratis-storage/project/issues/394

mulkieran:
Closes #740

### DIFF
--- a/src/cachedev.rs
+++ b/src/cachedev.rs
@@ -563,7 +563,7 @@ impl CacheDev {
         }
 
         let table = CacheDev::gen_default_table(&meta, &cache, &origin, cache_block_size);
-        let dev_info = device_create(dm, name, uuid, &table, create_options)?;
+        let dev_info = device_create(dm, name, uuid, &table, DmOptions::default(), create_options)?;
 
         Ok(CacheDev {
             dev_info: Box::new(dev_info),
@@ -599,7 +599,8 @@ impl CacheDev {
             device_match(dm, &dev, uuid)?;
             dev
         } else {
-            let dev_info = device_create(dm, name, uuid, &table, create_options)?;
+            let dev_info =
+                device_create(dm, name, uuid, &table, DmOptions::default(), create_options)?;
             CacheDev {
                 dev_info: Box::new(dev_info),
                 meta_dev: meta,
--- a/src/cachedev.rs
+++ b/src/cachedev.rs
@@ -546,6 +546,7 @@ impl DmDevice<CacheDevTargetTable> for CacheDev {
 impl CacheDev {
     /// Construct a new CacheDev with the given data and meta devs.
     /// Returns an error if the device is already known to the kernel.
+    #[allow(clippy::too_many_arguments)]
     pub fn new(
         dm: &DM,
         name: &DmName,
@@ -554,6 +555,7 @@ impl CacheDev {
         cache: LinearDev,
         origin: LinearDev,
         cache_block_size: Sectors,
+        create_options: DmOptions,
     ) -> DmResult<CacheDev> {
         if device_exists(dm, name)? {
             let err_msg = format!("cachedev {} already exists", name);
@@ -561,7 +563,7 @@ impl CacheDev {
         }
 
         let table = CacheDev::gen_default_table(&meta, &cache, &origin, cache_block_size);
-        let dev_info = device_create(dm, name, uuid, &table, DmOptions::default())?;
+        let dev_info = device_create(dm, name, uuid, &table, create_options)?;
 
         Ok(CacheDev {
             dev_info: Box::new(dev_info),
@@ -573,6 +575,7 @@ impl CacheDev {
     }
 
     /// Set up a cache device from the given metadata and data devices.
+    #[allow(clippy::too_many_arguments)]
     pub fn setup(
         dm: &DM,
         name: &DmName,
@@ -581,6 +584,7 @@ impl CacheDev {
         cache: LinearDev,
         origin: LinearDev,
         cache_block_size: Sectors,
+        create_options: DmOptions,
     ) -> DmResult<CacheDev> {
         let table = CacheDev::gen_default_table(&meta, &cache, &origin, cache_block_size);
         let dev = if device_exists(dm, name)? {
@@ -595,7 +599,7 @@ impl CacheDev {
             device_match(dm, &dev, uuid)?;
             dev
         } else {
-            let dev_info = device_create(dm, name, uuid, &table, DmOptions::default())?;
+            let dev_info = device_create(dm, name, uuid, &table, create_options)?;
             CacheDev {
                 dev_info: Box::new(dev_info),
                 meta_dev: meta,
@@ -762,7 +766,7 @@ pub fn minimal_cachedev(dm: &DM, paths: &[&Path]) -> CacheDev {
         meta_length,
         LinearDevTargetParams::Linear(meta_params),
     )];
-    let meta = LinearDev::setup(dm, &meta_name, None, meta_table).unwrap();
+    let meta = LinearDev::setup(dm, &meta_name, None, meta_table, DmOptions::default()).unwrap();
 
     let cache_name = test_name("cache-cache").expect("valid format");
     let cache_offset = meta_length;
@@ -773,7 +777,7 @@ pub fn minimal_cachedev(dm: &DM, paths: &[&Path]) -> CacheDev {
         cache_length,
         LinearDevTargetParams::Linear(cache_params),
     )];
-    let cache = LinearDev::setup(dm, &cache_name, None, cache_table).unwrap();
+    let cache = LinearDev::setup(dm, &cache_name, None, cache_table, DmOptions::default()).unwrap();
 
     let dev2_size = blkdev_size(&OpenOptions::new().read(true).open(paths[1]).unwrap()).sectors();
     let dev2 = Device::from(devnode_to_devno(paths[1]).unwrap().unwrap());
@@ -785,7 +789,8 @@ pub fn minimal_cachedev(dm: &DM, paths: &[&Path]) -> CacheDev {
         dev2_size,
         LinearDevTargetParams::Linear(origin_params),
     )];
-    let origin = LinearDev::setup(dm, &origin_name, None, origin_table).unwrap();
+    let origin =
+        LinearDev::setup(dm, &origin_name, None, origin_table, DmOptions::default()).unwrap();
 
     CacheDev::new(
         dm,
@@ -795,6 +800,7 @@ pub fn minimal_cachedev(dm: &DM, paths: &[&Path]) -> CacheDev {
         cache,
         origin,
         MIN_CACHE_BLOCK_SIZE,
+        DmOptions::default(),
     )
     .unwrap()
 }
--- a/src/core/dm.rs
+++ b/src/core/dm.rs
@@ -18,7 +18,7 @@ use crate::{
     core::{
         device::Device,
         deviceinfo::DeviceInfo,
-        dm_flags::DmFlags,
+        dm_flags::{DmCookie, DmFlags},
         dm_ioctl as dmi,
         dm_options::DmOptions,
         errors,
@@ -54,7 +54,11 @@ impl DmOptions {
         allowable_flags: DmFlags,
     ) -> DmResult<dmi::Struct_dm_ioctl> {
         let clean_flags = allowable_flags & self.flags();
-        let event_nr = u32::from(self.cookie().bits()) << 16;
+
+        // Every ioctl is a primary source of udev events; this flag should
+        // always be set when an ioctl is sent.
+        let event_nr =
+            u32::from(self.cookie().bits() | DmCookie::DM_UDEV_PRIMARY_SOURCE_FLAG.bits()) << 16;
         let mut hdr: dmi::Struct_dm_ioctl = Default::default();
 
         hdr.version[0] = dmi::DM_VERSION_MAJOR;
--- a/src/core/dm.rs
+++ b/src/core/dm.rs
@@ -58,7 +58,7 @@ impl DmOptions {
         // Every ioctl is a primary source of udev events; this flag should
         // always be set when an ioctl is sent.
         let event_nr =
-            u32::from(self.cookie().bits() | DmCookie::DM_UDEV_PRIMARY_SOURCE_FLAG.bits()) << 16;
+            u32::from((self.cookie() | DmCookie::DM_UDEV_PRIMARY_SOURCE_FLAG).bits()) << 16;
         let mut hdr: dmi::Struct_dm_ioctl = Default::default();
 
         hdr.version[0] = dmi::DM_VERSION_MAJOR;
--- a/src/lineardev.rs
+++ b/src/lineardev.rs
@@ -544,7 +544,8 @@ impl LinearDev {
             device_match(dm, &dev, uuid)?;
             dev
         } else {
-            let dev_info = device_create(dm, name, uuid, &table, create_options)?;
+            let dev_info =
+                device_create(dm, name, uuid, &table, DmOptions::default(), create_options)?;
             LinearDev {
                 dev_info: Box::new(dev_info),
                 table,
--- a/src/lineardev.rs
+++ b/src/lineardev.rs
@@ -532,6 +532,7 @@ impl LinearDev {
         name: &DmName,
         uuid: Option<&DmUuid>,
         table: Vec<TargetLine<LinearDevTargetParams>>,
+        create_options: DmOptions,
     ) -> DmResult<LinearDev> {
         let table = LinearDevTargetTable::new(table);
         let dev = if device_exists(dm, name)? {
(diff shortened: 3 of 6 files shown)


mulkieran:
Test failures due to an EBUSY.

mulkieran:
rebased.

mulkieran:
EBUSY, again.

jbaublitz (approved):

bmr-cymru:
I see we hit an EBUSY during the latest CI run:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Core(Ioctl(Some(DeviceInfo { version: Version { major: 4, minor: 45, patch: 0 }, data_size: 16384, data_start: 312, target_count: 0, open_count: 0, flags: (empty), event_nr: 4194304, dev: Device { major: 0, minor: 0 }, name: Some(DmNameBuf { inner: "cache-meta_dm-rs_test_delme" }), uuid: None }), EBUSY))', src/cachedev.rs:884:29
```

This is the `cache.teardown()` call at the end of `test_minimal_cache_dev`:

```rust
 817     // Test creating a minimal cache dev.
 818     // Verify that status method executes and gives reasonable values.
 819     fn test_minimal_cache_dev(paths: &[&Path]) {
...
 883 
 884         cache.teardown(&dm).unwrap();
 885     }
```

I have also been able to reproduce similar errors running the `sudo_test` target locally - they pop up in "random" places, and shift around with each run of the test suite. Common places to see them include:

- test thindev::tests::loop_test_setup_without_new ... FAILED
- test thinpooldev::tests::loop_test_status_noflush ... FAILED
- test thinpooldev::tests::loop_test_set_data ... FAILED
- test thinpooldev::tests::loop_test_set_meta ... FAILED
- test cachedev::tests::loop_test_origin_size_change ... FAILED

And likely others that I haven't observed yet.

From reviewing the possible locations that could return EBUSY from the corresponding device-mapper ioctl commands I had a suspicion that we were hitting this errno in `drivers/md/dm-ioctl.c::dev_remove()`:

```C
 952 static int dev_remove(struct file *filp, struct dm_ioctl *param, size_t param_size)
 953 {
 954         struct hash_cell *hc;
 955         struct mapped_device *md;
 956         int r;
 957         struct dm_table *t;
 ...
 970         /*
 971          * Ensure the device is not open and nothing further can open it.
 972          */
 973         r = dm_lock_for_deletion(md, !!(param->flags & DM_DEFERRED_REMOVE), false);
 974         if (r) {
 975                 if (r == -EBUSY && param->flags & DM_DEFERRED_REMOVE) {
 976                         up_write(&_hash_lock);
 977                         dm_put(md);
 978                         return 0;
 979                 }
 980                 DMDEBUG_LIMIT("unable to remove open device %s", hc->name);
 981                 up_write(&_hash_lock);
 982                 dm_put(md);
 983                 return r;
 984         }
 985 
...
1004 }
```
The function `dm_lock_for_deletion()` returns `EBUSY` if the `dm_open_count()` of the device is non-zero.

Recent Fedora and RHEL kernels are compiled with the DYNAMIC_DEBUG option, so DMDEBUG calls need to be enabled for the messages to appear on the console or in dmesg:

```
# echo 'file drivers/md/dm* +p' > /sys/kernel/debug/dynamic_debug/control
```

And re-running the test confirms that for each EBUSY we also log the "unable to remove open device" message:

```
[ 1688.260679] device-mapper: core: 253:3: suspending with flush
[ 1688.261014] device-mapper: table: loop0: error: dax unsupported by block device
[ 1688.263748] device-mapper: core: 253:4: suspending with flush
[ 1688.266161] device-mapper: ioctl: unable to remove open device data_dm-rs_test_delme <<<<<
[ 1688.266911] device-mapper: ioctl: unable to remove open device meta_dm-rs_test_delme <<<<<
[ 1688.301198] loop0: detected capacity change from 0 to 2097152
[ 1688.317537] device-mapper: table: loop0: error: dax unsupported by block device
[ 1688.317958] device-mapper: table: loop0: error: dax unsupported by block device
```

The way that the failures appear in different places, and change from run to run is highly suggestive of a race condition; I suspect that the EBUSY cases reflect situations where there was outstanding udev processing holding the device open at the time of the attempted removal.

bmr-cymru:
I hadn't considered that - I think it's worth evaluating and I'll take a look at that. I had been thinking of a couple of ways forward, both in terms of understanding the problem better and ultimately addressing it:

- Whether we can hook some monitoring into the tests to provide a "smoking gun" in terms of confirming who/what has the device open. I was thinking of this earlier in terms of needing to enable specific monitoring for a given named device (which is tricky since everything here is ephemeral and set up / torn down on the fly), but I think there's actually a way to get this for any device-mapper device and (hopefully) we'll be able to log the context of both the process invoking the open, and the device identity. I'll look into this at the same time as the retries.

- I think that to really prevent this from happening in general we need to implement the equivalent of `dm_udev_wait()` in devicemapper-rs so that we use the same SysV IPC synchronisation that the `dmsetup` calls in the generic udev rules are using. and then decide how best to plumb that into the Rust API. I had some discussions with John about possible ways to do that in our 1-1 but I'll bring it up at today's standup.

bmr-cymru:
I wanted to confirm the suspicion that the culprit here is udev to make sure we know what problem we're trying to solve. I put together a little systemtap script that monitors open/close on device-mapper devices, as well as EBUSY returns from the `dm_lock_for_deletion()` function. This initially threw up some confusing results yesterday (since it reported that the Rust devicemapper test binary was the culprit for some EBUSY returns!), so I've continued to enhance the script and follow the device-mapper code to get more insight into the problem and I think I understand what is happening now.

First it might help to give a little detail on how the `open_count` for device-mapper works. The counter is maintained as an atomic member (`open_count`) in the DM `mapped_device` struct (drivers/md/dm-core.h). The counter is incremented in `dm_blk_open()` and decremented in `dm_blk_close()`. These functions are set as the method pointers for the block device open and close hooks in `block_device_operations`:

```C
tatic const struct block_device_operations dm_blk_dops = {
        .submit_bio = dm_submit_bio,
        .poll_bio = dm_poll_bio,
        .open = dm_blk_open, <<<<<
        .release = dm_blk_close, <<<<<
        .ioctl = dm_blk_ioctl,
        .getgeo = dm_blk_getgeo,
        .report_zones = dm_blk_report_zones,
        .pr_ops = &dm_pr_ops,
        .owner = THIS_MODULE
};

static const struct block_device_operations dm_rq_blk_dops = {
        .open = dm_blk_open, <<<<<
        .release = dm_blk_close, <<<<<
        .ioctl = dm_blk_ioctl,
        .getgeo = dm_blk_getgeo,
        .pr_ops = &dm_pr_ops,
        .owner = THIS_MODULE
};
```

And these are only ever called via the function pointers in this struct: no kernel code ever calls them directly, but when device-mapper acquires a device-mapper device for use in a table, it calls the internal function `dm_table_get_device()`. This then proceeds through `dm_table_get_device()` -> `open_table_device()` -> `blkdev_get_by_dev()` -> `blkdev_get_whole()` and this function invokes the devices's open method via `disk->fops`:

```C
 660 static int blkdev_get_whole(struct block_device *bdev, fmode_t mode)
 661 {
 662         struct gendisk *disk = bdev->bd_disk;
 663         int ret;
 664 
 665         if (disk->fops->open) {
 666                 ret = disk->fops->open(bdev, mode); <<<<<
 667                 if (ret) {
 668                         /* avoid ghost partitions on a removed medium */
 669                         if (ret == -ENOMEDIUM &&
 670                              test_bit(GD_NEED_PART_SCAN, &disk->state))
 671                                 bdev_disk_changed(disk, true);
 672                         return ret;
 673                 }
 674         }
...
 682 }
```
So for our case there are two ways that the counter can be incremented:

- Externally: a process calls `open(2)` from userspace on the block device node or a symlink
- Internally: another device-mapper device opens the device by referencing it in its table

In both cases the DM device cannot be removed while the `open_count` is non-zero. 

The stap script works as follows:

- When a process invokes `dm_blk_open()` (either via `open(2)` or internally), we increment a counter indexed by a (pid, dm_name) tuple, where dm_name is the kernel name of the device, e.g. "dm-1".
- When a process invokes `dm_blk_close()` (via syscall or internally), we decrement the counter and delete the (pid, dm_name) entry if it reaches zero.

This means that at any time an error occurs we have a list of the open DM devices and the PIDs that opened them.

- Whenever a process calls `dm_lock_for_deletion()` (the function where the EBUSY originates), we check the return value and if it is EBUSY we log a message to the console like:

    pid 1234 dm_lock_for_deletion() -> EBUSY dm-1

And call `dmsetup ls --tree` to display the current DM device tree.

Then it will iterate over the list of openers and if the `dm_name` matches, we print the PID and dm_name:

    pid 53343 has dm-3 open

And for each PID seen we run `ps -o pid,args -p $PID` to display additional details of the process.

All of this is a little racey and non-deterministic; the pattern of errors varies from run-to-run of the test suite (from 0 to as many as 15 test failures in my runs), and it is possible that by the time our dmsetup and ps invocations are scheduled that the relevant devices or processes have already disappeared (with a greater number of errors the script has more work to do, making this more likely) but on a "good" run it provides a lot of insight into what's happening to the device reference counts, and why the EBUSY returns appear.

```
global openers
global verbose = 0
global printed

probe kernel.function("dm_lock_for_deletion").return {
    # EBUSY
    if ($return == -16) {
        dm_name = kernel_string($md->disk->disk_name)
        printf("pid %d dm_lock_for_deletion() -> EBUSY %s\n", pid(), dm_name)
        system("dmsetup ls --tree")
        foreach ([p, d] in openers) {
            if (d == dm_name) {
                printf("pid %d has %s open\n", p, d)
                if (!printed[p]) {
                    system(sprintf("ps -o pid,args -p %d", p))
                    printed[p] = 1
                }
            }
        }
        delete printed
    }
}

probe kernel.function("dm_blk_open") {
    dm_name = kernel_string($bdev->bd_disk->disk_name)
    openers[pid(), dm_name]++
    if (verbose)
        printf("dm_blk_open() %s %d %s\n", dm_name, pid(), execname())
}

probe kernel.function("dm_blk_close") {
    dm_name = kernel_string($disk->disk_name)
    openers[pid(), dm_name]--
    if (!openers[pid(), dm_name])
        delete openers[pid(), dm_name]
    if (verbose)
        printf("dm_blk_close() %s %d %s\n", dm_name, pid(), execname())
}
```

The script is run with the `stap` command and requires the `kernel-debuginfo` packages to be installed:

```
# stap ~/dmrs741.stp 
WARNING: confusing usage, value is captured as @entry($md->disk->disk_name) in .return probe [man stapprobes] RETURN PROBES: identifier '$md' at /root/dmrs741.stp:6:33
 source:         dm_name = kernel_string($md->disk->disk_name)
                                         ^
warning: the compiler differs from the one used to build the kernel
  The kernel was built by: gcc (GCC) 12.0.1 20220413 (Red Hat 12.0.1-0)
  You are using:           gcc (GCC) 12.1.1 20220507 (Red Hat 12.1.1-1)
```

The WARNING is harmless and can be ignored: it's indicating that we're accessing a function argument during the function return probe. This may give misleading results in some cases since the value may change while the function executes but in this case we know that the `mapped_device` is still valid since we are returning `EBUSY`.

Once started the script will wait for events to occur - I'll post the output and analysis in a follow-up comment.

bmr-cymru:
All the systemtap probes run concurrently and asynchronously, so it's possible for the output to get a bit mixed up in the terminal. This is the log from a run which had two test failures and it shows what's going on quite well:

```
test cachedev::tests::loop_test_suspend ... FAILED
test thindev::tests::loop_test_setup_without_new ... FAILED
```

```
pid 53973 dm_lock_for_deletion() -> EBUSY dm-4
pid 53978 has dm-4 open
pid 53973 dm_lock_for_deletion() -> EBUSY dm-2
pid 53973 has dm-2 open
pid 53973 dm_lock_for_deletion() -> EBUSY dm-1
pid 53973 has dm-1 open
pid 53973 dm_lock_for_deletion() -> EBUSY dm-3
pid 53973 has dm-3 open
cache-cache_dm-rs_test_delme (253:2)
 └─ (7:0)
cache-meta_dm-rs_test_delme (253:1)
 └─ (7:0)
cache-origin_dm-rs_test_delme (253:3)
 └─ (7:1)
fedora_fedora-root (253:0)
 └─ (252:2)
cache-cache_dm-rs_test_delme (253:2)
 └─ (7:0)
cache-meta_dm-rs_test_delme (253:1)
 └─ (7:0)
cache-origin_dm-rs_test_delme (253:3)
 └─ (7:1)
fedora_fedora-root (253:0)
 └─ (252:2)
cache-meta_dm-rs_test_delme (253:1)
 └─ (7:0)
cache-origin_dm-rs_test_delme (253:3)
 └─ (7:1)
fedora_fedora-root (253:0)
 └─ (252:2)
cache-meta_dm-rs_test_delme (253:1)
 └─ (7:0)
cache-origin_dm-rs_test_delme (253:3)
 └─ (7:1)
fedora_fedora-root (253:0)
 └─ (252:2)
    PID COMMAND
  53978 /usr/lib/systemd/systemd-udevd
    PID COMMAND
  53973 /home/bmr/src/git/devicemapper-rs/target/debug/deps/devicemapper-7524c876aa27eca2
    PID COMMAND
  53973 /home/bmr/src/git/devicemapper-rs/target/debug/deps/devicemapper-7524c876aa27eca2
    PID COMMAND
  53973 /home/bmr/src/git/devicemapper-rs/target/debug/deps/devicemapper-7524c876aa27eca2
pid 53973 dm_lock_for_deletion() -> EBUSY dm-3
pid 53974 has dm-3 open
pid 53973 dm_lock_for_deletion() -> EBUSY dm-2
pid 53973 has dm-2 open
pid 53973 dm_lock_for_deletion() -> EBUSY dm-1
pid 53973 has dm-1 open
data_dm-rs_test_delme (253:2)
 └─ (7:0)
fedora_fedora-root (253:0)
 └─ (252:2)
meta_dm-rs_test_delme (253:1)
 └─ (7:0)
data_dm-rs_test_delme (253:2)
 └─ (7:0)
fedora_fedora-root (253:0)
 └─ (252:2)
meta_dm-rs_test_delme (253:1)
 └─ (7:0)
fedora_fedora-root (253:0)
 └─ (252:2)
meta_dm-rs_test_delme (253:1)
 └─ (7:0)
    PID COMMAND
    PID COMMAND
  53974 /usr/lib/systemd/systemd-udevd
  53973 /home/bmr/src/git/devicemapper-rs/target/debug/deps/devicemapper-7524c876aa27eca2
    PID COMMAND
  53973 /home/bmr/src/git/devicemapper-rs/target/debug/deps/devicemapper-7524c876aa27eca2
pid 53973 dm_lock_for_deletion() -> EBUSY dm-2
pid 53973 has dm-2 open
pid 53973 dm_lock_for_deletion() -> EBUSY dm-1
pid 53973 has dm-1 open
fedora_fedora-root (253:0)
 └─ (252:2)
meta_dm-rs_test_delme (253:1)
 └─ (7:0)
fedora_fedora-root (253:0)
 └─ (252:2)
meta_dm-rs_test_delme (253:1)
 └─ (7:0)
    PID COMMAND
  53973 /home/bmr/src/git/devicemapper-rs/target/debug/deps/devicemapper-7524c876aa27eca2
    PID COMMAND
  53973 /home/bmr/src/git/devicemapper-rs/target/debug/deps/devicemapper-7524c876aa27eca2
```

There are three PIDs interacting in this log, two instances of `systemd-udevd` and one of the devicemapper-rs test suite:

- 53973 devicemapper-7524c876aa27eca2 - the test suite
- 53974 /usr/lib/systemd/systemd-udevd
- 53978 /usr/lib/systemd/systemd-udevd

Initially the test suite (53973) tries to remove dm-4 but it is opened by udev:

```
pid 53973 dm_lock_for_deletion() -> EBUSY dm-4
pid 53978 has dm-4 open
```

In this case dm-4 is the cache device that is consuming the other three DM devices:

- dm-1 (cache-meta_dm-rs_test_delme)
- dm-2 (cache-cache_dm-rs_test_delme)
- dm-3 (cache-origin_dm-rs_test_delme)

The test suite then proceeds to try to remove each of these devices in sequence - but this fails, since the devices are still opened by dm-4 (which we could not remove because it was open by udev).

The iterative cleanup in the test suite via `testing::test_lib::clean_up() -> dm_test_devices_remove()` mostly deals with this, but occasionally will report:

  Some test-generated DM devices remaining: ...
  Failed to ensure removal of all test-generated DM devices

This also explains why very occasionally the `make sudo_test` completes with zero errors, but the systemtap script detects one or more EBUSY: these EBUSY are being seen by the cleanup code, which suppresses the error and silently pushes the device back onto the list of remaining devices. It's only if we hit an EBUSY in a test case that these get logged, for e.g. by the `cache.teardown(&dm).unwrap();` at the end of `test_minimal_cache_dev()`.

So there are two distinct causes of the EBUSY errors, one external and one internal to the test suite:

- A race with systemd's udev child processes opening devices
- devicemapper-rs attempting to tear down a stack out-of-order while top-level devices still have underlying layers open

mulkieran:
As an aside, because of bz https://bugzilla.redhat.com/show_bug.cgi?id=1506287, the test code does not use deferred removal, which in theory would be cleaner and might avoid the EBUSY's from the test code.

mulkieran:
Superceded by #774.

PR #770, opened 2022-07-18
Set a per-command ioctl version in device-mapper header

bmr-cymru:
Adopt the approach taken by libdevmapper and use a table to determine
the correct ioctl interface version to specify for each possible ioctl
command.

This allows devicemapper-rs to function with a broader range of kernels
since it will only request a higher ioctl version number from the kernel
when issuing a command that requires it and avoids failures in CI test
environments where devicemapper-rs may be built with newer device-mapper
features than are supported by the running kernel.

### DIFF
--- a/src/core/dm.rs
+++ b/src/core/dm.rs
@@ -55,12 +55,12 @@ impl DmOptions {
     ) -> DmResult<dmi::Struct_dm_ioctl> {
         let clean_flags = allowable_flags & self.flags();
         let event_nr = u32::from(self.cookie().bits()) << 16;
-        let mut hdr: dmi::Struct_dm_ioctl = Default::default();
-
-        hdr.flags = clean_flags.bits();
-        hdr.event_nr = event_nr;
-
-        hdr.data_start = size_of::<dmi::Struct_dm_ioctl>() as u32;
+        let mut hdr: dmi::Struct_dm_ioctl = devicemapper_sys::dm_ioctl {
+            flags: clean_flags.bits(),
+            event_nr,
+            data_start: size_of::<dmi::Struct_dm_ioctl>() as u32,
+            ..Default::default()
+        };
 
         if let Some(id) = id {
             match id {


src/core/dm.rs
```diff
@@ -121,6 +150,12 @@ impl DM {
         // to copy the hdr into v, and later to update the
         // possibly-modified hdr.
 
+        if let Some(cmd_version) = ioctl_version_for_cmd(ioctl) {
+            hdr.version[0] = cmd_version.0;
+            hdr.version[1] = cmd_version.1;
+            hdr.version[2] = cmd_version.2;
+        }
```
bmr-cymru, line 157: This keeps the old behaviour of using the compiled-against libdevmapper ioctl version for unknown commands, but I'm leaning toward thinking that we should instead treat that as an error condition?

bmr-cymru:
It's easy enough to trim the table down by special-casing the original v4 commands, but it feels a bit like deckchair shuffling:

```rust
const CMD_VERSIONS: [(u32, u32, u32); 5] = [
    (4, 1, 0),
    (4, 2, 0),
    (4, 6, 0),
    (4, 36,0),
    (4, 41, 0),
];

fn ioctl_version_for_cmd(cmd: u8) -> Option<(u32, u32, u32)> {
    if cmd <= dmi::DM_TABLE_STATUS_CMD as u8 {
        Some((4, 0, 0))
    } else if cmd <= dmi::DM_GET_TARGET_VERSION_CMD as u8 {
        let idx = cmd as usize - dmi::DM_LIST_VERSIONS_CMD as usize;
        Some(CMD_VERSIONS[idx])
    } else {
        None
    }
}
```

It would be helpful if libdevmapper had boundary constants for the ioctl commands and I'm wondering if the table couldn't be generated automatically from the generated bindings somehow? I'm fairly limited in bingden knowledge but I'll take a look at this now.

bmr-cymru:
The CI tests have revealed another drawback with the simple approach:

```
   Compiling cexpr v0.6.0
   Compiling nix v0.24.2
   Compiling devicemapper-sys v0.1.2 (/__w/devicemapper-rs/devicemapper-rs/devicemapper-rs-sys)
   Compiling devicemapper v0.32.0 (/__w/devicemapper-rs/devicemapper-rs)
error[E0425]: cannot find value `DM_GET_TARGET_VERSION_CMD` in module `dmi`
   --> src/core/dm.rs:102:20
    |
102 |     if cmd <= dmi::DM_GET_TARGET_VERSION_CMD as u8 {
    |                    ^^^^^^^^^^^^^^^^^^^^^^^^^ not found in `dmi`
```

The copy of libdevmapper used in this CI environment pre-dates the introduction of this ioctl command:

```
-=================================
+Version 1.02.165 - 23rd October 2019
+====================================
   Add support for DM_DEVICE_GET_TARGET_VERSION.
   Add debug of dmsetup udevcomplete with hexa print DM_COOKIE_COMPLETED.
   Fix versioning of dm_stats_create_region and dm_stats_create_region.
```

Which means that devicemapper-rs cannot use the symbolic name without triggering a build error (A `DM_IOCTL_CMD_MAX` would be helpful here!).

src/core/dm.rs
bmr-cymru: v2 now treats unknown command values as errors.

src/core/dm.rs
```diff
@@ -121,6 +149,17 @@ impl DM {
...
+            hdr.version[0] = ioctl_version.0;
+            hdr.version[1] = ioctl_version.1;
+            hdr.version[2] = ioctl_version.2;
+        } else {
+            return Err(DmError::Core(errors::Error::InvalidArgument(format!(
+                "Unknown ioctl command: {}",
+                ioctl
+            ))));
+        }
+
```
mulkieran, line 162: I was going to ask you not to separate the comment chunk from its neighbor, but you can just get rid of the whole comment, both chunks, instead. The whole comment here is obsolete or duplicated elsewhere.

mulkieran (changes requested):
I'm definitely happy with this change overall, but have a few comments and suggestions.

mulkieran (approved):
Looks good!

jbaublitz (changes requested):
@mulkieran I requested that @bmr-cymru try taking a feature flagging approach for constants in devicemapper-rs like we do in libcryptsetup-rs based on the version. We should be able to both use the header versions to determine feature flags while also using the table to use the minimum accepted version like `libdm` to avoid kernel/userspace header mismatch problems.

src/core/dm.rs
```diff
@@ -77,6 +77,34 @@ impl DmOptions {
...
+];
+
+fn ioctl_to_version(ioctl: u8) -> Option<(u32, u32, u32)> {
+    if ioctl as u32 <= DM_MAX_V4_IOCTL_CMD {
+        Some((4, 0, 0))
+    } else if ioctl as u32 <= DM_MAX_IOCTL_CMD {
+        let idx = ioctl as u32 - (DM_MAX_V4_IOCTL_CMD + 1);
+        Some(CMD_VERSIONS[idx as usize])
+    } else {
+        None
```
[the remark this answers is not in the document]
bmr-cymru: Thanks! I actually had the function returning just a tuple at one point, but it didn't occur to me how to handle the unreachable branch.

src/core/dm.rs
bmr-cymru: Removed in updated branch.

src/core/dm.rs
```diff
@@ -647,6 +641,7 @@ impl DM {
 
     /// Returns a list of each loaded target type with its name, and
     /// version broken into major, minor, and patchlevel.
+    #[cfg(devicemapper46supported)]
```
mulkieran, line 644: I think this should properly be "41" not "46".

mulkieran (changes requested):
One thing I just noticed, otherwise looks good.

src/core/dm.rs
bmr-cymru: Thank you!

bmr-cymru:
I've rebased again to fix the problem @mulkieran pointed out - originally i was only generating feature flags as far back as devicemapper46supported. When I amended build.rs to also generate the older flags I neglected to fix up the feature flags in dm.rs to use the correct versioning.

mulkieran (approved):

jbaublitz (approved):

PR #774, opened 2022-07-26
Add udev semaphore synchronization for uevent-generating commands

bmr-cymru:
This branch adds support for udev synchronization using SysV semaphores and the device-mapper ioctl/uevent "cookie" mechanism.

The branch adds a new sub-module, `dm_udev_sync` that wraps the low-level semaphore operations and provides an interface to `core::dm`.

All udev synchronization is carried out insde `DM::do_ioctl()`. This centralises and simplifies handling ioctl errors (which require the udev semaphore to be cleaned up before propagating the error) and minimises the changes needed to synchronize with uevents.

This branch also adds basic logging infrastructure to the crate, using the log facade and the env_logger crate: this was helpful in developing and debugging the udev interactions and may be useful for debugging other problems involving device-mapper.

### DIFF
--- a/src/core/dm.rs
+++ b/src/core/dm.rs
@@ -21,6 +21,7 @@ use crate::{
         dm_flags::DmFlags,
         dm_ioctl as dmi,
         dm_options::DmOptions,
+        dm_udev_sync::{UdevSync, UdevSyncAction},
         errors,
         types::{DevId, DmName, DmNameBuf, DmUuid},
         util::{
@@ -54,7 +55,7 @@ impl DmOptions {
         allowable_flags: DmFlags,
     ) -> DmResult<dmi::Struct_dm_ioctl> {
         let clean_flags = allowable_flags & self.flags();
-        let event_nr = u32::from(self.udev_flags().bits()) << 16;
+        let event_nr = self.udev_flags().bits() << dmi::DM_UDEV_FLAGS_SHIFT;
         let mut hdr: dmi::Struct_dm_ioctl = devicemapper_sys::dm_ioctl {
             flags: clean_flags.bits(),
             event_nr,
@@ -120,6 +121,10 @@ impl DM {
         hdr.version[1] = ioctl_version.1;
         hdr.version[2] = ioctl_version.2;
 
+        // Begin udev sync transaction and set DM_UDEV_PRIMARY_SOURCE_FLAG
+        // if ioctl command generates uevents.
+        let sync = UdevSync::begin(hdr, ioctl)?;
+
         let data_size = cmp::max(
             MIN_BUF_SIZE,
             size_of::<dmi::Struct_dm_ioctl>() + in_data.map_or(0, |x| x.len()),
@@ -148,6 +153,8 @@ impl DM {
             if let Err(err) = unsafe {
                 convert_ioctl_res!(nix_ioctl(self.file.as_raw_fd(), op, buffer.as_mut_ptr()))
             } {
+                // Cancel udev sync and clean up semaphore
+                sync.cancel();
                 return Err(DmError::Core(errors::Error::Ioctl(
                     op as u8,
                     DeviceInfo::new(*hdr).ok().map(Box::new),
@@ -174,6 +181,8 @@ impl DM {
 
         let data_end = cmp::max(buffer_hdr.data_size, buffer_hdr.data_start);
 
+        // Synchronize with udev event processing
+        sync.end(buffer_hdr.flags)?;
         Ok((
             DeviceInfo::try_from(*buffer_hdr)?,
             buffer[buffer_hdr.data_start as usize..data_end as usize].to_vec(),
--- a/src/core/dm.rs
+++ b/src/core/dm.rs
@@ -319,6 +319,7 @@ impl DM {
             Self::hdr_set_uuid(&mut hdr, uuid)?;
         }
 
+        debug!("Creating device {} (uuid={:?})", name, uuid);
         self.do_ioctl(dmi::DM_DEV_CREATE_CMD as u8, &mut hdr, None)
             .map(|(hdr, _)| hdr)
     }
@@ -333,6 +334,7 @@ impl DM {
     pub fn device_remove(&self, id: &DevId<'_>, options: DmOptions) -> DmResult<DeviceInfo> {
         let mut hdr = options.to_ioctl_hdr(Some(id), DmFlags::DM_DEFERRED_REMOVE)?;
 
+        debug!("Removing device {}", id);
         self.do_ioctl(dmi::DM_DEV_REMOVE_CMD as u8, &mut hdr, None)
             .map(|(hdr, _)| hdr)
     }
@@ -358,6 +360,7 @@ impl DM {
         let mut hdr = options.to_ioctl_hdr(None, DmFlags::DM_UUID)?;
         Self::hdr_set_name(&mut hdr, old_name)?;
 
+        debug!("Renaming device {} to {}", old_name, new);
         self.do_ioctl(dmi::DM_DEV_RENAME_CMD as u8, &mut hdr, Some(&data_in))
             .map(|(hdr, _)| hdr)
     }
@@ -391,6 +394,12 @@ impl DM {
             DmFlags::DM_SUSPEND | DmFlags::DM_NOFLUSH | DmFlags::DM_SKIP_LOCKFS,
         )?;
 
+        let action = if options.flags().contains(DmFlags::DM_SUSPEND) {
+            "Suspending"
+        } else {
+            "Resuming"
+        };
+        debug!("{} device {}", action, id);
         self.do_ioctl(dmi::DM_DEV_SUSPEND_CMD as u8, &mut hdr, None)
             .map(|(hdr, _)| hdr)
     }
@@ -401,6 +410,7 @@ impl DM {
     pub fn device_info(&self, id: &DevId<'_>) -> DmResult<DeviceInfo> {
         let mut hdr = DmOptions::default().to_ioctl_hdr(Some(id), DmFlags::empty())?;
 
+        debug!("Retrieving info for {}", id);
         self.do_ioctl(dmi::DM_DEV_STATUS_CMD as u8, &mut hdr, None)
             .map(|(hdr, _)| hdr)
     }
@@ -420,6 +430,7 @@ impl DM {
     ) -> DmResult<(DeviceInfo, Vec<(u64, u64, String, String)>)> {
         let mut hdr = options.to_ioctl_hdr(Some(id), DmFlags::DM_QUERY_INACTIVE_TABLE)?;
 
+        debug!("Waiting on event for {}", id);
         let (hdr_out, data_out) = self.do_ioctl(dmi::DM_DEV_WAIT_CMD as u8, &mut hdr, None)?;
 
         let status = DM::parse_table_status(hdr.target_count, &data_out)?;
@@ -507,6 +518,7 @@ impl DM {
         // Flatten targets into a buf
         let data_in = cursor.into_inner();
 
+        debug!("Loading table \"{:?}\" for {}", targets, id);
         self.do_ioctl(dmi::DM_TABLE_LOAD_CMD as u8, &mut hdr, Some(&data_in))
             .map(|(hdr, _)| hdr)
     }
@@ -515,6 +527,7 @@ impl DM {
     pub fn table_clear(&self, id: &DevId<'_>) -> DmResult<DeviceInfo> {
         let mut hdr = DmOptions::default().to_ioctl_hdr(Some(id), DmFlags::empty())?;
 
+        debug!("Clearing inactive dable for {}", id);
         self.do_ioctl(dmi::DM_TABLE_CLEAR_CMD as u8, &mut hdr, None)
             .map(|(hdr, _)| hdr)
     }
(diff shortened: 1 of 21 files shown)


mulkieran:
@bmr-cymru Plz put up a stratisd PR that uses these changes in the manner intended. You can just edit that PRs Cargo.toml to point the devicemapper-rs dependency at this PR.

bmr-cymru:
Pulling this back into draft for now to fix some problems I found while testing with stratis.

bmr-cymru:
Rebased to current master and added a pair of commits with the changes to move `DM_UDEV_PRIMARY_SOURCE_FLAG` handling into the `UdevSync::begin()` method.

src/core/dm_udev_sync.rs
```diff
@@ -0,0 +1,506 @@
...
+                Err(udev_sync_error_from_os())
+            }
+            _ => Ok(()),
+        }
+    }
+
+    #[derive(Debug)]
+    pub struct UdevSync {
+        cookie: u32,
+        semid: i32,
```
jbaublitz, line 277: Should this be an `Option` instead? It looks like you're using -1 as a `None` stand in.

src/core/dm_udev_sync.rs
```diff
@@ -0,0 +1,506 @@
...
+        pub semmni: ::std::os::raw::c_int,
+        pub semmns: ::std::os::raw::c_int,
+        pub semmnu: ::std::os::raw::c_int,
+        pub semmsl: ::std::os::raw::c_int,
+        pub semopm: ::std::os::raw::c_int,
+        pub semume: ::std::os::raw::c_int,
+        pub semusz: ::std::os::raw::c_int,
+        pub semvmx: ::std::os::raw::c_int,
+        pub semaem: ::std::os::raw::c_int,
+    }
```
jbaublitz, line 84: I have two questions here. Why is this only for musl? I see these structures in the glibc headers too. I'm wondering if I'm misunderstanding something. Do you think it would make more sense to use bindgen here too?

jbaublitz (commented):
I did a first pass and have some suggestions. I want to dig a little bit deeper past basic understanding and style comments.

src/core/dm_udev_sync.rs
```diff
@@ -0,0 +1,506 @@
...
+        val: c_int,
+        buf: &'a semid_ds,
+        array: &'a c_ushort,
+        __buf: &'a seminfo,
+    }
+
+    // Missing libc SysV IPC constants from /usr/include/linux/sem.h
+    const GETVAL: i32 = 12;
+    const SETVAL: i32 = 16;
+    const SEM_INFO: i32 = 19;
```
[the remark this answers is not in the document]
bmr-cymru: Would that be a new crate, like devicemapper-rs-sys? I could add it to the existing bindings generation but it doesn't seem like a good fit with the ioctl / libdevmapper constants?

src/core/dm_udev_sync.rs
bmr-cymru: I'm not entirely sure why this difference exists in the `libc` crate; there is a sub-module, `gnu`, that contains `struct seminfo` (in gnu/mod.rs), and `struct semid_ds` (in gnu/b{32,64}/mod.rs). There is a separate `musl` sub-module that is compiled for musl targets that does not include these definitions:

src/unix/linux_like/linux/mod.rs
```rust
cfg_if! {
    if #[cfg(target_env = "uclibc")] {
        mod uclibc;
        pub use self::uclibc::*;
    } else if #[cfg(target_env = "musl")] {
        mod musl;
        pub use self::musl::*;
    } else if #[cfg(target_env = "gnu")] {
        mod gnu;
        pub use self::gnu::*;
    }
}
```

Looking at the commit that introduced this for `gnu`:

```
commit 775f0f20d37e56e615a88471678784ab7abf52ca
Author: David Carlier <devnexen@gmail.com>
Date:   Sun Jun 20 11:00:44 2021 +0100

    linux adding semid_ds struct and seminfo incidently.
    
    close #2002
```

And looking at the issue that it closed, rust-lang/libc#2002 it may just be an omission that the definitions are missing from musl?

This is a bit of a new type of problem for me, so I'm open to suggestions on the best way to handle it - I'm happy to file an issue to add these in the libc crate upstream. There are issues with some libc updates for musl due to the musl approach of using sanitised kernel headers (which may be out of date and lacking recent definitions - see e.g. rust-lang/libc#1571), but I don't think that will be an issue for the SysV semaphore interface.

src/core/dm_udev_sync.rs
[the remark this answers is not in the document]
bmr-cymru: I've added a set of commits to introduce a new `sysvsem-sys` crate, and to use bindgen for both the missing consts and (for musl builds) the two missing struct definitions.

This seems to work well and cleans up the use statements and definitions at the top of dm_udev_sync.

src/core/dm_udev_sync.rs
```diff
@@ -0,0 +1,459 @@
...
+
+        fn cancel(self) {
+            if !self.is_active() {
+                return;
+            }
+            let semid = self.semid.unwrap();
+            debug!("Canceling {:?}", self);
+            if let Err(err) = notify_sem_destroy(self.cookie, semid) {
+                error!("Failed to clean up notification semaphore: {}", err);
+            }
```
jbaublitz, line 304: I think avoid early return when not necessary is usually better. Something like this:

```rust
if self.is_active() {
   let semid = ...
}
```

is more idiomatic.

jbaublitz (changes requested):
Okay, I think other than these comments, this looks great!

src/core/dm_udev_sync.rs
```diff
@@ -0,0 +1,459 @@
...
+
+    use super::UdevSyncAction;
+
+    #[repr(C)]
+    union semun<'a> {
+        val: c_int,
+        buf: &'a semid_ds,
+        array: &'a c_ushort,
+        __buf: &'a seminfo,
+    }
```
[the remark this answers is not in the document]
bmr-cymru: The `semun` union is one of the weirder parts of the SysV interface - I believe originally the `semctl` function prototype used a void pointer for the 4th argument:

```
int semctl(int semid, int semnum, int cmd, void *);
```

Rather than the variadic function that is now standardised:

```
int semctl(int semid, int semnum, int cmd, ...);
```

The union is not defined in the SysV header files - rather it's shown in the manual page and standards docs that must be defined by the calling program:

```
       This function has three or four arguments, depending on cmd. 
       When there are four, the fourth has the type union semun. 
       The calling program must define this union as follows:

           union semun {
               int              val;    /* Value for SETVAL */
               struct semid_ds *buf;    /* Buffer for IPC_STAT, IPC_SET */
               unsigned short  *array;  /* Array for GETALL, SETALL */
               struct seminfo  *__buf;  /* Buffer for IPC_INFO
                                           (Linux-specific) */
           };
```

If we include that definition in the header that is passed to bindgen for the SysV pieces it produces the following Rust, which seems simpler and sufficient for our uses:

```Rust
#[repr(C)]
#[derive(Copy, Clone)]
pub union semun {
    pub val: ::std::os::raw::c_int,
    pub buf: *mut semid_ds,
    pub array: *mut ::std::os::raw::c_ushort,
    pub __buf: *mut seminfo,
}
```

src/core/dm.rs
```diff
@@ -40,6 +43,12 @@ const DM_CTL_PATH: &str = "/dev/device-mapper";
 /// Start with a large buffer to make BUFFER_FULL rare. Libdm does this too.
 const MIN_BUF_SIZE: usize = 16 * 1024;
 
+/// Number of device remove retry attempts
+const DM_REMOVE_RETRIES: usize = 25;
```
mulkieran, line 47: 25 seems like so many to me; if there is a reference or any other reason for this number it's worthwile putting it here.

src/core/dm_flags.rs
```diff
@@ -53,23 +53,23 @@ bitflags! {
     /// https://sourceware.org/git/?p=lvm2.git;a=blob;f=libdm/libdevmapper.h#l3627
     /// for complete information about the meaning of the flags.
     #[derive(Default)]
-    pub struct DmCookie: dmi::__u16 {
+    pub struct DmUdevFlags: u32 {
```
mulkieran, line 56: Try dropping ```identity_op``` clippy allow, it _should_ be obsolete with this change.

mulkieran (changes requested):
Two minor requests...

src/core/dm.rs
bmr-cymru: I tend to agree - the values currently used (including the delay time) are taken directly from libdevmapper, but there doesn't appear to be any real rationale to the specific values other than to try to really make sure the removal succeeds in the face of possible interference from outside processes like udev.

In my own testing, which is the most empirical data I have for this I see the vast majority of retries succeed on the first retry, a small number on the second, and very occasionally a third or fourth so my instinct is that we can reduce this quite substantially assuming we do decide to go ahead with the retries.

This is on my list for today's standup as unfortunately the last CI failure that would require the mechanism is still unresolved.

src/core/dm_flags.rs
bmr-cymru: Gotcha, thanks! Looking into/testing this..

src/core/dm_udev_sync.rs
```diff
@@ -0,0 +1,459 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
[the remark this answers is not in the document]
bmr-cymru: I've added additional comments to both the low level and `UdevSync` interfaces now. If they look OK to you I'll mark this as resolved?

src/core/dm_udev_sync.rs
```diff
@@ -0,0 +1,459 @@
...
+        }
+    }
+
+    fn notify_sem_create() -> DmResult<(u32, i32)> {
+        let cookie_pair = retry(NoDelay.take(4), gen_cookie);
+        if let Err(err) = cookie_pair {
+            error!("Failed to generate udev notification semaphore: {}", err);
+            return Err(DmError::Core(errors::Error::UdevSync(err.to_string())));
+        }
+        let (cookie, semid) = cookie_pair.unwrap();
```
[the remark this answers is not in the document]
bmr-cymru: Done, thanks!

src/core/dm_udev_sync.rs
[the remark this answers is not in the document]
bmr-cymru: The reference has been eliminated now that we are using bindgen-generated versions of these types.

src/core/dm_udev_sync.rs
bmr-cymru: Oh I had missed this! One moment...

bmr-cymru: This one should be addressed now.

mulkieran:
@bmr-cymru Note that bindgen dependency specification is now 0.63, so your fix at https://github.com/rust-lang/rust-bindgen/pull/2247 is available. This may not change what you want to do with the source at all, but you will want to fix up the commit message.

src/core/dm.rs
bmr-cymru: The current branch uses up to 5 retries and this seems adequate in all my testing.

bmr-cymru:
Good point. I'll update the commit to drop that mention at least - I can test removing the allowlist although in some ways I think it's desirable even with the fix for the `await` keyword. It keeps the bindings to the minimum set that we need and avoids pulling in unwanted types and interfaces.

bmr-cymru:
I think it's best to keep the allow listing - this avoids conflicts between the types in the generated bindings that result from the same types being exposed in both the device-mapper and SysV semaphore headers:

```

I'll edit the commit message accordingly.
error[E0428]: the name `__uid_t` is defined multiple times
    --> /home/breeves/src/git/devicemapper-rs/target/debug/build/devicemapper-sys-ab734666951c00f5/out/sem-bindings.rs:14:1
     |
14   | pub type __uid_t = ::std::os::raw::c_uint;
     | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `__uid_t` redefined here
     |
    ::: /home/breeves/src/git/devicemapper-rs/target/debug/build/devicemapper-sys-ab734666951c00f5/out/dm-bindings.rs:1294:1
     |
1294 | pub type __uid_t = ::std::os::raw::c_uint;
     | ------------------------------------------ previous definition of the type `__uid_t` here
     |
     = note: `__uid_t` must be defined only once in the type namespace of this module

error[E0428]: the name `__gid_t` is defined multiple times
    --> /home/breeves/src/git/devicemapper-rs/target/debug/build/devicemapper-sys-ab734666951c00f5/out/sem-bindings.rs:15:1
     |
15   | pub type __gid_t = ::std::os::raw::c_uint;
     | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `__gid_t` redefined here
     |
    ::: /home/breeves/src/git/devicemapper-rs/target/debug/build/devicemapper-sys-ab734666951c00f5/out/dm-bindings.rs:1295:1
     |
1295 | pub type __gid_t = ::std::os::raw::c_uint;
     | ------------------------------------------ previous definition of the type `__gid_t` here
     |
     = note: `__gid_t` must be defined only once in the type namespace of this module
```

mulkieran (changes requested):
Last batch of minor suggestions.

mulkieran (approved):

jbaublitz (approved):

mulkieran:
@bmr-cymru Ok! It looks like we're good to go. Let me know when you've squashed all the commits that need to be squashed, and then I'll merge the PR and make a new release.

bmr-cymru:
@mulkieran I've squashed all the fixup commits into their parents and tidied the branch up a little. It's still 12 commits but I think it makes a logical sequence of changes. Let me know if you'd like me to squash more aggressively.

PR #806, opened 2022-12-12
Correct buffer management in do_ioct and its calling methods

mulkieran:


src/core/dm.rs
```diff
@@ -102,90 +103,106 @@ impl DM {
...
             if len == u32::MAX as usize {
                 return Err(DmError::Core(errors::Error::IoctlResultTooLarge));
             }
-            v.resize((len as u32).saturating_mul(2) as usize, 0);
-
-            // v.resize() may move the buffer if the requested increase doesn't fit in continuous
-            // memory.  Update hdr to the possibly new address.
-            hdr = unsafe { &mut *(v.as_mut_ptr() as *mut dmi::Struct_dm_ioctl) };
-            hdr.data_size = v.len() as u32;
+            buffer.resize((len as u32).saturating_mul(2) as usize, 0);
```
jbaublitz, line 172: As I mentioned previously, I don't think doubling the size makes sense when devicemapper explicitly returns the size required to fit the message. We could simply use the output `data_size` which I think is simpler and will result in a maximum of two iterations.

src/core/dm.rs
```diff
@@ -102,90 +103,106 @@ impl DM {
...
         }
 
-        // hdr possibly modified so copy back
-        hdr_slc.clone_from_slice(&v[..hdr.data_start as usize]);
+        let data_end = cmp::max(buffer_hdr.data_size, buffer_hdr.data_start);
```
jbaublitz, line 175: Should this be `data_size + data_start`? I'm not sure `max()` does what we want in the case where `data_size < data_start` but still nonzero. This seems like it would truncate the data.

jbaublitz (changes requested):
Two comments, but other than that, this looks good.

src/core/dm.rs
mulkieran: I don't think we can rely on that. Take a look at the implementation of ```target_message```. We see:

```
        if (r == 1) {
                param->flags |= DM_DATA_OUT_FLAG;
                if (dm_message_test_buffer_overflow(result, maxlen))
                        param->flags |= DM_BUFFER_FULL_FLAG;
                else
                        param->data_size = param->data_start + strlen(result) + 1;
                r = 0;
        }
```

src/core/dm.rs
mulkieran: I don't know why data_size is sometimes set to less than data_start, so max seems necessary. data_size always includes the length of the hdr, which is data_start.

mulkieran: Ok. I believe it's clear why ```data_size``` is sometimes set to less than ```data_start```.  Before the selected ioctl computations are done, ```data_size``` is set to the offset to the start of the ```data``` field in in the dm_ioctl struct, which is 7 bytes less than the size of the whole struct. This is born out by tests. The message stratisd sends always sets ```data_start``` to the size of the whole struct. I do not believe that the data field is every actually used, it seems to function just as padding.

jbaublitz: This makes sense!

src/core/dm.rs
jbaublitz: It looks like I was totally wrong after looking at the kernel code. It appears it only sets the `data_size` parameter not only there but *everywhere* on success and on failure, it isn't touched. Resolving this.

jbaublitz (approved):
This looks reasonable to me!

mulkieran:
Rewrite buffer management commit msg was slightly out of date wrt. the commit.

bgurney-rh (approved):

PR #820, opened 2023-01-19
Increase nix dependency lower bound to 0.26.0

mulkieran:
Related https://github.com/stratis-storage/project/issues/570


mulkieran:
Expect some Fedora dependency checks to fail.

PR #830, opened 2023-02-15
Re add android

mulkieran:


mulkieran:
Still encountering problems with Android repos.

jbaublitz:
@flxo Do you have any recommendations for a container we can use to test Android?

flxo:
@mulkieran Not a good one ;-)

I used the [cross](https://github.com/cross-rs/cross) images with some [custom extensions](https://github.com/esrlabs/northstar/blob/964f87b8861611c49a190841aa7fabfd69f386bf/Cross.toml#L4). My problem here is that latest bingen [isn't compatible](https://github.com/esrlabs/northstar/actions/runs/3823564394/jobs/6504866977) with the base images in cross 0.2 and I await `cross 0.3` updated base images.

There are circle ci images I have never used.

flxo:
@mulkieran I just stumbled across the `libc` [android container](https://github.com/rust-lang/libc/blob/master/ci/docker/aarch64-linux-android/Dockerfile). In theory checking the same as `libc` is probably a good idea since I've seen a lot of android builds fail - not because the of the crate but of stuff missing in `libc` for android`.

PR #859, opened 2023-05-26
Test for presence of udev daemon and prevent stratisd-min from starting if it is absent

bmr-cymru:
Test whether the udev daemon appears to be running (by checking the udevd control socket path at /run/udev/control) and refuse to create devices if it is missing.

This change causes stratisd or stratisd-min to exit immediately and log the fact that udev is not running:

```
# /usr/libexec/stratisd-min
[2023-05-26T15:45:02Z ERROR stratisd::stratis::run] The devicemapper thread exited with an error: DM error: DM Core error: failed to perform udev sync operation: Udev daemon is not running: unable to create devices.; shutting down stratisd...
Error: "DM error: DM Core error: failed to perform udev sync operation: Udev daemon is not running: unable to create devices."
```

```
# systemctl start stratisd
# systemctl status stratisd
× stratisd.service - Stratis daemon
     Loaded: loaded (/usr/lib/systemd/system/stratisd.service; enabled; preset: enabled)
    Drop-In: /usr/lib/systemd/system/service.d
             └─10-timeout-abort.conf
     Active: failed (Result: exit-code) since Fri 2023-05-26 11:47:31 EDT; 2s ago
   Duration: 12ms
       Docs: man:stratisd(8)
    Process: 1339 ExecStart=/usr/libexec/stratisd --log-level debug (code=exited, status=1/FAILURE)
   Main PID: 1339 (code=exited, status=1/FAILURE)
        CPU: 30ms

May 26 11:47:31 localhost.localdomain stratisd[1339]: [2023-05-26T15:47:31Z INFO  stratisd::engine::strat_engine::liminal::device_info] Device information Stratis device description: Strati>
May 26 11:47:31 localhost.localdomain stratisd[1339]: [2023-05-26T15:47:31Z INFO  stratisd::engine::strat_engine::liminal::device_info] Device information Stratis device with logical device>
May 26 11:47:31 localhost.localdomain stratisd[1339]: [2023-05-26T15:47:31Z DEBUG stratisd::engine::strat_engine::backstore::crypt::shared] Stratis LUKS2 token: {"activation_name":"stratis->
May 26 11:47:31 localhost.localdomain stratisd[1339]: [2023-05-26T15:47:31Z DEBUG stratisd::engine::strat_engine::backstore::crypt::shared] Stratis LUKS2 token: {"activation_name":"stratis->
May 26 11:47:31 localhost.localdomain stratisd[1339]: [2023-05-26T15:47:31Z INFO  stratisd::engine::strat_engine::liminal::liminal] Attempt to set up pool failed, but it may be possible to >
May 26 11:47:31 localhost.localdomain stratisd[1339]: [2023-05-26T15:47:31Z ERROR stratisd::stratis::run] The devicemapper thread exited with an error: DM error: DM Core error: failed to pe>
May 26 11:47:31 localhost.localdomain systemd[1]: Started stratisd.service - Stratis daemon.
May 26 11:47:31 localhost.localdomain stratisd[1339]: DM error: DM Core error: failed to perform udev sync operation: Udev daemon is not running: unable to create devices.
May 26 11:47:31 localhost.localdomain systemd[1]: stratisd.service: Main process exited, code=exited, status=1/FAILURE
May 26 11:47:31 localhost.localdomain systemd[1]: stratisd.service: Failed with result 'exit-code'.
```

The check applies to all DM ioctls so the daemon exits when the devicemapper thread attempts to check the device-mapper version. The alternative (only checking for ioctls that actually require udev synchronisation) places the check too late, since crypt devices will have already been set up. Since cryptsetup uses libdevmapper it will fall back to directly creating device nodes in /dev/mapper and this is not handled by the rollback code, leaving users to manually clean up these entries.

### DIFF
--- a/src/core/dm_udev_sync.rs
+++ b/src/core/dm_udev_sync.rs
@@ -31,9 +31,11 @@ pub mod sync_semaphore {
         IPC_RMID,
     };
 
+    use nix::unistd::{access, AccessFlags};
+
     use rand::Rng;
     use retry::{delay::NoDelay, retry, OperationResult};
-    use std::io;
+    use std::{io, path::Path};
 
     use crate::core::sysvsem::seminfo;
 
@@ -97,6 +99,15 @@ pub mod sync_semaphore {
                 }
             }
         }
+    }
+
+    const UDEV_SOCKET_PATH: &str = "/run/udev/control";
+
+    fn udev_running() -> bool {
+        matches!(
+            access(Path::new(UDEV_SOCKET_PATH), AccessFlags::F_OK),
+            Ok(())
+        )
     }
 
     /// Allocate or retrieve a SysV semaphore set identifier
@@ -291,6 +302,12 @@ pub mod sync_semaphore {
         /// Allocate a SysV semaphore according to the device-mapper udev cookie
         /// protocol and set the initial state of the semaphore counter.
         fn begin(hdr: &mut dmi::Struct_dm_ioctl, ioctl: u8) -> DmResult<Self> {
+            if !udev_running() {
+                return Err(DmError::Core(errors::Error::UdevSync(
+                    "Udev daemon is not running: unable to create devices.".to_string(),
+                )));
+            }
+
             match ioctl as u32 {
                 dmi::DM_DEV_REMOVE_CMD | dmi::DM_DEV_RENAME_CMD | dmi::DM_DEV_SUSPEND_CMD
                     if *SYSV_SEM_SUPPORTED && (hdr.flags & DmFlags::DM_SUSPEND.bits()) == 0 => {}


mulkieran (approved):

jbaublitz:
Just one comment here. Is lazy-static the best option? We've bumped into things like this for Clevis support where if the state can change, we have to consider whether we care or not. If stratisd is not restarted, udev synchronization will be turned off permanently even if systemd-udevd is restarted and the queue becomes active. @mulkieran What are your thoughts on this?

bmr-cymru:
I think for Stratis it's preferable in this situation that we prevent the daemon from starting up - the problem is that libcryptsetup uses libdevmapper which has the library fall-back code to directly create device nodes in /dev/mapper. In my tests the rollback code in stratisd is not able to tear this down successfully so we are left in a funny state with a half-built stack. The only way I could find to recover from this was to manually remove the crypt devices with dmsetup.

Exposing this in devicemapper and having the daemon check `udev_running()` (or whatever) might be a better way of handling this. The check for the socket is very cheap so this could actually go and test each time rather than using the lazy_static.

mulkieran:
> I think for Stratis it's preferable in this situation that we prevent the daemon from starting up - the problem is that libcryptsetup uses libdevmapper which has the library fall-back code to directly create device nodes in /dev/mapper. In my tests the rollback code in stratisd is not able to tear this down successfully so we are left in a funny state with a half-built stack. The only way I could find to recover from this was to manually remove the crypt devices with dmsetup.
> 
> Exposing this in devicemapper and having the daemon check `udev_running()` (or whatever) might be a better way of handling this. The check for the socket is very cheap so this could actually go and test each time rather than using the lazy_static.


I agree with the Stratis decision. I'm only wondering if there's a change in stratisd that would make that decision more robust to future, unrelated code changes in stratisd. If stratisd, for some reason, moved the version check out of the dm setup code, then a side-effect of that would be that stratisd would not be brought down when the setup code is run.

bmr-cymru:
I'll update this to drop the lazy_static! and just call the `udev_running()` function each time - the `access()` check is very cheap so I think there's no concern about performance here and as John points out it prevents us from getting out of sync if the state of udev changes while we are running.

mulkieran (approved):

jbaublitz (approved):

PR #890, opened 2023-11-07
Downgrade DM debugging messages to trace level

bmr-cymru:
Commit 66e1fdf2 added logging to the DM module using the debug! macro.

In practice these messages can be very noisy, especially for e.g. in the Stratis test suite (which calls the status() method on thin pool devices). This leads to lots of messages like the following:

[2023-11-06T16:01:05Z DEBUG devicemapper::core::dm] Retrieving table status for stratis-1-private-50db9929d265403b820bb155c8f5ae89-thinpool-pool

Running the Stratis test suite results in hundreds of these messages being logged at debug level.

Downgrade the logging calls in the DM module from debug! to trace! to quiet this down.

### DIFF
--- a/src/core/dm.rs
+++ b/src/core/dm.rs
@@ -384,7 +384,7 @@ impl DM {
         match retry_with_index(
             Fixed::from_millis(DM_REMOVE_MSLEEP_DELAY).take(DM_REMOVE_RETRIES - 1),
             |i| {
-                debug!("Device remove attempt {} of {}", i, DM_REMOVE_RETRIES);
+                trace!("Device remove attempt {} of {}", i, DM_REMOVE_RETRIES);
                 self.try_device_remove(id, options)
             },
         ) {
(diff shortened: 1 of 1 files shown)


src/core/dm.rs
```diff
@@ -586,7 +586,7 @@ impl DM {
     pub fn table_clear(&self, id: &DevId<'_>) -> DmResult<DeviceInfo> {
         let mut hdr = DmOptions::default().to_ioctl_hdr(Some(id), DmFlags::empty())?;
 
-        debug!("Clearing inactive dable for {}", id);
+        trace!("Clearing inactive dable for {}", id);
```
mulkieran, line 589: Might as well fix typo also: ^dable^table

mulkieran (approved):
This is fine w/ me, although if you wanted to leave the bigger actions, create, rename, remove at debug level, I would be fine w/ that also.

mulkieran (changes requested):
Forgot about the typo...definitely should fix that.

src/core/dm.rs
bmr-cymru: Well spotted, thanks!

jbaublitz (approved):

bmr-cymru:
> This is fine w/ me, although if you wanted to leave the bigger actions, create, rename, remove at debug level, I would be fine w/ that also.

I think that would make sense. Those operations should be considerably less frequent in any case. I'll update the PR.

bmr-cymru:
I've revised this to leave the create, remove, rename, suspend/resume and target message logs at debug level with the rest downgraded to trace.

mulkieran (approved):

PR #978, opened 2025-04-28
devicemapper-rs-sys: Use lowest supported Rust as bindgen Rust target

mulkieran:


mulkieran:
Note that generated code is identical.

PR #992, opened 2025-07-09
Export lineardev::FeatureArg as FlakeyFeatureArg

frankdavid:


mulkieran (changes requested):
It needs formatting, but after that, should be mergeable.

PR #1004, opened 2025-10-05
Use dev_t as return value

taamrooz:
This uses the `dev_t` type rather than `u64` for returning the device number from the device node.

`dev_t` isn't always `u64`, so this addresses an issue that could occur on platforms where `dev_t` would be `u32` for example.


jbaublitz:
@taamrooz If the tests pass, I believe we should be all set!

